PR #10, opened 2019-11-23
Add opus support

nvllsvm:
Mimetype defined in RFC7845 https://tools.ietf.org/html/rfc7845#section-9

Confirmed working with Audinaut.


sentriz:
great thank you :)

sentriz:
currently building on docker hub

https://hub.docker.com/repository/registry-1.docker.io/sentriz/gonic/builds/69e9888c-6c6a-4a61-9132-598eb60c1a4d

PR #35, opened 2020-02-13
WIP: Add track transcoding support

spijet:
This PR contains my local edits to gonic that I made to implement track transcoding. Currently, it has a hard dependency on `ffmpeg` and features 4 encoding profiles — MP3 128k CBR and Opus 96k VBR, with or without forced ReplayGain (useful for mobile clients).

To reduce load on server when listening to same tracks over and over, I also added file cache. It's fairly stupid ATM, as it doesn't manage old and incorrect entries (i.e. when FFmpeg crashed half-way through encoding a track), so any additions are welcome.

I also didn't bother with adding proper tests, since I lack needed experience (and I'm a lazy butt :P).


sentriz:
Hi! Thank you very much for doing this :) I had a quick look - I'm happy this can be done is so few lines.

I'm currently in the middle of a college work "sprint" which ends on Tuesday, so I can do a proper look then

spijet:
Of course, take your time. :)
I'll add more stuff to this PR in the meantime, if I come up with anything else.

server/ctrlsubsonic/handlers_cache.go
```diff
@@ -0,0 +1,155 @@
...
+	ENC_PROFILES = map[string]*encoderProfile {
+		"mp3"    : {  "mp3", "128k", []string{"-c:a", "libmp3lame"}                    , false },
+		"mp3_rg" : {  "mp3", "128k", []string{"-c:a", "libmp3lame"}                    , true  },
+		"opus"   : { "opus",  "96k", []string{"-c:a", "libopus", "-vbr", "constrained"}, false },
+		"opus_rg": { "opus",  "96k", []string{"-c:a", "libopus", "-vbr", "constrained"}, true  },
+	}
+	BUF_LEN = 4096
+)
+
+func StreamTrack(w http.ResponseWriter, r *http.Request, trackPath string, client string, cachePath string) {
```
sentriz: i think this can be unexported since it's only used in this package (as well as other funcs in this file)

server/ctrlbase/ctrl.go
```diff
@@ -47,6 +47,7 @@ func statusToBlock(code int) string {
 type Controller struct {
 	DB          *db.DB
 	MusicPath   string
+	CachePath   string
```
sentriz: i think this can be moved to the Controller in `ctrlsubsonic` instead, since it's only used there (instead of here where both the admin page and subsonic use it)

spijet: How do I do it properly? AFAICT, at the moment `ctrlsubsonic` gets spawned by `server` with all options copied from `ctrlbase`.

server/ctrlsubsonic/handlers_cache.go
spijet: Should be fixed now (if I got the "exported/unexported" idea right). :)

server/ctrlbase/ctrl.go
sentriz: oops you'd need to pass an arg to SetupSubsonic. nevermind then. i need to refactor server.go anyway so i'll do it then

spijet:
@sentriz, It seems that my fork also has a changed file in `assets/assets_gen.go`, which seems to be auto-generated (I didn't edit it myself). Should I include it in this PR?

spijet:
Also, it may be better to allow user to set up encoding profiles via Gonic's UI (similar to OG Subsonic), than to hardcode them. The only problem that may arise is that Subsonic spawns a shell process to run the command (to allow multi-step conversions, e.g. `ffmpeg ... | opusenc ...`), while this PR runs a single process (which MAY be a shell... :stuck_out_tongue:).

spijet:
> ...and then linted with `golangci-lint run` from the root dir

Done (mostly). As for now, `golangci-lint` wants me to replace all `http.ResponseWriter` and `os.File` type references with `io.Writer`, and I'm not sure if that's wise.

spijet:
Added support for user-specified bitrate.
Some clients allow the user to limit the bitrate used for streaming, using the `maxBitRate` URI parameter. I thought that it could be useful for mobile apps (e.g. Soundwaves), so I added a check if this parameter is set.

If it's set to a value lower than one in selected encoding profile, it'll be passed to `ffmpeg` as the target bitrate instead.

sentriz:
nice one. may I ask why you don't like the `io.Writer` idea?

spijet:
I don't have enough experience with Go to be 100% sure it'll work properly. :smile:
Besides, `writeCmdOutput()` uses `http.Flusher`, so I thought that it'd be better to leave it as it is.

But if it's OK to use `io.Writer`, I can do it. :)

By the way, I compared `writeCmdOutput()` and `copyCmdOutput()` again, and it seems that the latter actually gives me lower TTFB (i.e. gonic starts streaming encoded output faster) and allows `ffmpeg` to perform the encoding a bit faster, probably because `io.Copy` just tries to spit the output as fast as possible, leaving stuff like output buffering to the OS? I'm actually surprised that it's faster, given that `writeCmdOutput()` force-flushes the HTTP response every 4k bytes, which should be pretty quick. In my case, `write...()` gives me TTFB of 220-240ms, while `copy...()` gives 110-180ms.
*More testing needed!* :rofl:

sentriz:
whoops. didn't see this as marked ready. can have a look on tuesday :)

spijet:
I'll try to "backport" your commits from `transcoding_update` branch until then. :)

spijet:
Just updated my branch to current master. Hopefully nothing's broken :D
I'll also pull commits 76e07a9ed69b35829f0ecee9e11edcd0fafda038 and 7fc4566fda7556f47d42d2f2f39855b916b25ee2 for completeness sake.

sentriz:
hey. haven't forgot about this :) currently working on a ui for a user to select a profile on a per client basis. this also means that transcoding won't be mandatory

spijet:
Oops, now there's a conflict in three files:
```
assets/assets_gen.go
assets/pages/home.tmpl
assets/static/main.css
```

spijet:
I solved the conflict in `home.tmpl` and `main.css`, but `assets_gen.go` is a bit harder to figure out. :)

sentriz:
for that file just reset/checkout the file then `./_do_gen_assets` and commit. should be fine

spijet:
Aaaand done! :)

sentriz:
:D thanks again for the feature

PR #42, opened 2020-03-02
add getGenres support

### LINKED ISSUE #34: getGenres Support
It seems that Gonic does not support the `getGenres` endpoint. Is this in the roadmap? (I know you are busy, I just graduated and I know the final year pains.)

This is affecting a user of my Subsonic client:
https://gitlab.com/sumner/sublime-music/issues/109#note_284391742
(arguably, my client should be more resilient to not having certain endpoints missing, and that's a goal I have in the future, but I just wanted to make you aware on the server-side.)

Duncaen:
I never used gomigrate, so I'm not sure if this is correct.

This also adds a decoded name column to the genres table, I thought there was a indexed genre but there isn't I can remove the column if you think its not necessary.

closes #34

### DIFF
--- a/server/ctrlsubsonic/handlers_common.go
+++ b/server/ctrlsubsonic/handlers_common.go
@@ -288,18 +288,27 @@ func (c *Controller) ServeGetSong(r *http.Request) *spec.Response {
 
 func (c *Controller) ServeGetRandomSongs(r *http.Request) *spec.Response {
 	params := r.Context().Value(CtxParams).(params.Params)
-	// TODO: add genre restraint here
 	var tracks []*db.Track
-	c.DB.DB.
+
+	q := c.DB.DB.Joins("JOIN albums ON tracks.album_id=albums.id").
 		Limit(params.GetIntOr("size", 10)).
-		Where(
-			"albums.tag_year BETWEEN ? AND ?",
-			params.GetIntOr("fromYear", 1800),
-			params.GetIntOr("toYear", 2200)).
-		Joins("JOIN albums ON tracks.album_id=albums.id").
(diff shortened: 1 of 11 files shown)


sentriz:
hi! thank you very much for this - looks very good. 
as far as i can remember the "u dec" columns are only used for searching (so you can look up artist/tracks/albums with non latin chars with just latin char equivalents) so i think it's safe to remove them for genres

https://github.com/sentriz/gonic/blob/master/server/ctrlsubsonic/handlers_by_tags.go#L157-L210

Duncaen:
I removed the decoded field from the genre table, added a `getSongsByGenre` view and added genre parameter support to the `getRandomSongs` view.

One additional thing I could add is add a genre parameter to `spec.NewTrackByTags` so that all the views that return `TrackChild` can fill in the genre field.

sentriz:
cool thank you :) i can have a proper look and merge when my next college work "sprint" ends. shouldn't be long

Duncaen:
Ok nice, I rebased on master and squashed the fixup commits, this is ready for review.

server/ctrlsubsonic/handlers_common.go
```diff
@@ -288,18 +288,27 @@ func (c *Controller) ServeGetSong(r *http.Request) *spec.Response {
 
 func (c *Controller) ServeGetRandomSongs(r *http.Request) *spec.Response {
```
sentriz: thanks for fixing my `year` laziness here :)

sentriz:
sorry for the delay. nothing much i would change here. thank you for the feature :)

PR #43, opened 2020-03-02
spec: add IgnoredArticles attribute to Index responses

Duncaen:
This attribute is required as per spec and keeping it empty,
complies the spec as long as articles are not ignored while indexing.
In the future, this could maybe be a user setting that could be
supported.

With this change, the aurial client works now.

* https://shrimpza.github.io/aurial/
* https://github.com/shrimpza/aurial


sentriz:
very nice thank you

PR #58, opened 2020-04-17
Add support for the jukebox endpoint

AlexKraak:
This supports most of jukeboxControl.view as far as i can tell. Things
seem to be playing ok without freaking out

I've also only tested it a little bit with ultrasonic but it does
appear to be working pretty well


sentriz:
thank you ali!!!

spijet:
I'm a bit late to the party, but…

Instead of using separate libraries for FLAC and MP3 formats (and, AFAICT, missing out on other possible formats), why not try to use FFmpeg to decode the track to raw wave format and then play it? Or it can be used as a fallback for formats that aren't supported with native libs.

sentriz:
not a bad idea. what do you think @AlexKraak ?

spijet:
Maybe it's even possible to link with `libavcodec` (FFmpeg's main library) directly instead of calling the binary, but it'd be harder to implement and probably not as secure (since there were some RCE vulnerabilities in `libavcodec` in the past :P).

Also, a small rant: 48kHz (`48000`) is not an optimal sampling rate in most cases, as most (if not all) mainstream music formats (like CD Audio and most WWW releases) are 16bit/44.1kHz. 
48kHz makes sense for Opus, DVD Audio and maybe some other formats, and in rare cases when hardware is incapable of 44.1kHz playback and resampling the audio in software yields much better quality than it does in hardware (hello, SoundBlaster cards).

PR #67, opened 2020-05-15
Added the instructions for the use with systemd

IUCCA:


sentriz:
very cool thank you!

contrib/config
```diff
@@ -0,0 +1,7 @@
+music-path        
+cache-path        /tmp/gonic_cache
+db-path           var/gonic/gonic.db
```
sentriz, line 3: should this instead be `/var/gonic/gonic.db`?

README.md
```diff
@@ -64,6 +64,51 @@ services:
...
+sudo nano /var/gonic/config
+```
+
+Setup systemd service:
+
+```bash
+sudo wget https://raw.githubusercontent.com/sentriz/gonic/master/contrib/gonic.service -O /etc/systemd/system/gonic.service
+sudo systemctl daemon-reload
+sudo systemctl start gonic
+sudo systemctl enable gonic
```
sentriz, line 107: might be a litter nice instead:
```suggestion
sudo systemctl enable --now gonic
```

contrib/config
IUCCA: Oh yes, your right.

README.md
```diff
@@ -64,6 +64,50 @@ services:
...
+```bash
+go get go.senan.xyz/gonic/cmd/gonic
+```
+
+Add a gonic user and create a directory for the server:
+
+```bash
+sudo mkdir -p /var/gonic/
+sudo adduser --disabled-login --gecos "" gonic
+sudo mv go/* /var/gonic/
```
sentriz, line 89: sorry, another question. what does this `mv` do?

IUCCA: go get install the files by deafult in the folder go in your homedir, this mv changed the path from ~/go/ to /var/gonic.

sentriz: ahh i see. well just incase the user has other go software installed, it might be better to change to something like `mv go/bin/gonic /var/gonic`? then change the `ExecStart` below to remove the `bin/`? 
what do you think? this way we only copy the binary

PR #69, opened 2020-05-19
Bump oto to allow building on OpenBSD

qbit:
Hi, newer versions of oto allow for building on FreeBSD and OpenBSD.

This PR bumps oto and allows gonic to run on OpenBSD!

`go mod tidy` was run.


sentriz:
thanks! :)

PR #71, opened 2020-06-04
Update README formatting and badges

sumnerevans:
Update README formatting and made the issues and pull request badges actually go to the Issues and Pull requests pages.

Also removed leading space at the front of all of the lines in the README because I have OCD.


sentriz:
thanks!

PR #75, opened 2020-06-16
Make duration adhere to the spec

### LINKED ISSUE #74: Duration returned by playlists is not an integer
I found an issue with the implementation of playlists.

The spec specifies the type of `duration` to be `int`, but Gonic returns a string.

```
        <xs:attribute name="duration" type="xs:int" use="required"/>     <!--Added in 1.8.0-->
```

This breaks Playlists in the latest Sublime Music.

sumnerevans:
Closes #74 

I think this works, but I am no expert in Go.


sumnerevans:
Ok, I tested this and it seems to work when I run Gonic locally.

sentriz:
thanks!! reminds me I need to make that return the right thing, not 1

PR #76, opened 2020-06-16
Fix grammar in error message

edwargix:


edwargix:
Let me know if you'd rather have the commit message start with a lowercase letter as seems to be the convention in this repo.

sentriz:
thanks :)

PR #85, opened 2020-07-24
Add duration and track_count to getArtist, getAlbum and getAlbumList2

Duncaen:


server/ctrlsubsonic/handlers_by_tags.go
```diff
@@ -139,7 +145,7 @@ func (c *Controller) ServeGetAlbumListTwo(r *http.Request) *spec.Response {
 	// TODO: think about removing this extra join to count number
 	// of children. it might make sense to store that in the db
 	q.
-		Select("albums.*, count(tracks.id) child_count").
+		Select("albums.*, count(tracks.id) child_count, sum(tracks.id) duration").
```
sentriz, line 148: should this be `sum(tracks.length)` instead?

server/ctrlsubsonic/handlers_by_tags.go
```diff
@@ -54,7 +54,11 @@ func (c *Controller) ServeGetArtist(r *http.Request) *spec.Response {
 	}
 	artist := &db.Artist{}
 	c.DB.
-		Preload("Albums").
+		Preload("Albums", func(db *gorm.DB) *gorm.DB {
+			return db.Select("*, count(sub.id) child_count, sum(sub.length) duration").
```
sentriz, line 58: sorry I have very silly formatting rules, can we get a newline after the `db.`?

sentriz:
also thinking to myself. a lot of joins are happening now to get the duration of things. probably doesn't make sense to calculate this everytime. but its probably ok for now

Duncaen:
Updated the commit to include the suggested changes.


There is already a TODO comment about the joins for the track count, but I'm not really familiar with gorm so I decided to use the queries for now.

sentriz:
brill thank you :) i'll make a release soon

sentriz:
added in v0.10.2

PR #86, opened 2020-07-24
Empty getArtistInfo2 response if lastfm_api_key is not set

Duncaen:
DSub will error out in Artists views if browse by tags is enabled
and getArtistInfo2 returns an error.


sentriz:
thanks, makes sense

PR #87, opened 2020-07-24
add album cover scaling and caching

Duncaen:
This uses a default width of 600 which is what airsonic uses.
As format png is used which could be changed to jpg, I don't really have a preference.
The default compression level is used, this could be changed to better speed or better compression.

### DIFF
--- a/cmd/gonic/main.go
+++ b/cmd/gonic/main.go
@@ -60,7 +60,7 @@ func main() {
 			log.Fatalf("couldn't create cache path: %v\n", err)
 		}
 	}
-	coverCachePath := path.Join(*cachePath, "covers")
+	coverCachePath := path.Join(*cachePath, coverCachePrefix)
 	if _, err := os.Stat(coverCachePath); os.IsNotExist(err) {
 		if err := os.MkdirAll(coverCachePath, os.ModePerm); err != nil {
 			log.Fatalf("couldn't create cover cache path: %v\n", err)
--- a/cmd/gonic/main.go
+++ b/cmd/gonic/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path"
 	"regexp"
 	"time"
 
@@ -15,6 +16,11 @@ import (
 	"go.senan.xyz/gonic/server"
 	"go.senan.xyz/gonic/server/db"
 	"go.senan.xyz/gonic/version"
+)
+
+const (
+	cleanTimeDuration = 10 * time.Minute
+	coverCachePrefix  = "covers"
 )
 
 func main() {
@@ -45,12 +51,19 @@ func main() {
 	set.VisitAll(func(f *flag.Flag) {
 		log.Printf("    %-15s %s\n", f.Name, f.Value)
 	})
+	//
 	if _, err := os.Stat(*musicPath); os.IsNotExist(err) {
 		log.Fatal("please provide a valid music directory")
 	}
 	if _, err := os.Stat(*cachePath); os.IsNotExist(err) {
 		if err := os.MkdirAll(*cachePath, os.ModePerm); err != nil {
 			log.Fatalf("couldn't create cache path: %v\n", err)
+		}
+	}
+	coverCachePath := path.Join(*cachePath, "covers")
+	if _, err := os.Stat(coverCachePath); os.IsNotExist(err) {
+		if err := os.MkdirAll(coverCachePath, os.ModePerm); err != nil {
+			log.Fatalf("couldn't create cover cache path: %v\n", err)
 		}
 	}
 	db, err := db.New(*dbPath)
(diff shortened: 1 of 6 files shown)


Duncaen:
Images are only downscaled to the width requested by size or 600 as the default, the aspect ratio is preserved.

sentriz:
nice thanks! just did some testing and works very well :) i was _going_ to say how maybe it would be better to do this during the import/scan phase. but yeah looks like the client can request any size so this makes sense :) 

how would you feel about this patch? all it does is split the handler into two smaller funcs so that the stuff under the `stat` is less indented (also removes some logging because i think it could be a bit noisy when scrolling though pages of albums)

```diff
diff --git a/server/ctrlsubsonic/handlers_raw.go b/server/ctrlsubsonic/handlers_raw.go
index 2e2cb62..e995b3b 100644
--- a/server/ctrlsubsonic/handlers_raw.go
+++ b/server/ctrlsubsonic/handlers_raw.go
@@ -61,6 +61,43 @@ const (
 	coverCacheFormat = "png"
 )
 
+func coverGetPath(dbc *db.DB, musicPath string, id int) (string, error) {
+	folder := &db.Album{}
+	err := dbc.DB.
+		Select("id, left_path, right_path, cover").
+		First(folder, id).
+		Error
+	if gorm.IsRecordNotFoundError(err) {
+		return "", fmt.Errorf("could not find a cover with that id")
+	}
+	if folder.Cover == "" {
+		return "", fmt.Errorf("no cover found for that folder")
+	}
+	return path.Join(
+		musicPath,
+		folder.LeftPath,
+		folder.RightPath,
+		folder.Cover,
+	), nil
+}
+
+func coverScaleAndSave(absPath, cachePath string, size int) error {
+	src, err := imaging.Open(absPath)
+	if err != nil {
+		return fmt.Errorf("resizing `%s`: %v", absPath, err)
+	}
+	width := size
+	if width > src.Bounds().Dx() {
+		// don't upscale images
+		width = src.Bounds().Dx()
+	}
+	err = imaging.Save(imaging.Resize(src, width, 0, imaging.Lanczos), cachePath)
+	if err != nil {
+		return fmt.Errorf("caching `%s`: %v", cachePath, err)
+	}
+	return nil
+}
+
 func (c *Controller) ServeGetCoverArt(w http.ResponseWriter, r *http.Request) *spec.Response {
 	params := r.Context().Value(CtxParams).(params.Params)
 	id, err := params.GetID("id")
@@ -68,51 +105,26 @@ func (c *Controller) ServeGetCoverArt(w http.ResponseWriter, r *http.Request) *s
 		return spec.NewError(10, "please provide an `id` parameter")
 	}
 	size := params.GetOrInt("size", coverDefaultSize)
-	cacheFile := path.Join(
+	cachePath := path.Join(
 		c.CachePath,
 		fmt.Sprintf("%s-%d.%s", id.String(), size, coverCacheFormat),
 	)
-	_, err = os.Stat(cacheFile)
-	if os.IsNotExist(err) {
-		log.Printf("serving cover `%s`: cache [%s/%d] miss!\n",
-			cacheFile, coverCacheFormat, size)
-		folder := &db.Album{}
-		err = c.DB.
-			Select("id, left_path, right_path, cover").
-			First(folder, id.Value).
-			Error
-		if gorm.IsRecordNotFoundError(err) {
-			return spec.NewError(10, "could not find a cover with that id")
-		}
-		if folder.Cover == "" {
-			return spec.NewError(10, "no cover found for that folder")
-		}
-		absPath := path.Join(
-			c.MusicPath,
-			folder.LeftPath,
-			folder.RightPath,
-			folder.Cover,
-		)
-		src, err := imaging.Open(absPath)
+	_, err = os.Stat(cachePath)
+	switch {
+	case os.IsNotExist(err):
+		coverPath, err := coverGetPath(c.DB, c.MusicPath, id.Value)
 		if err != nil {
-			log.Printf("resizing cover `%s`: error: %v\n", absPath, err)
+			return spec.NewError(10, "couldn't find cover %q: %v", id, err)
+		}
+		if err := coverScaleAndSave(coverPath, cachePath, size); err != nil {
+			log.Printf("error scaling cover: %v", err)
 			return nil
 		}
-		width := size
-		if width > src.Bounds().Dx() {
-			// don't upscale images
-			width = src.Bounds().Dx()
-		}
-		err = imaging.Save(imaging.Resize(src, width, 0, imaging.Lanczos), cacheFile)
-		if err != nil {
-			log.Printf("caching cover `%s`: error: %v\n", cacheFile, err)
-			return nil
-		}
-	} else {
-		log.Printf("serving cover `%s`: cache [%s/%d] hit!\n",
-			cacheFile, coverCacheFormat, size)
+	case err != nil:
+		log.Printf("error stating `%s`: %v", cachePath, err)
+		return nil
 	}
-	http.ServeFile(w, r, cacheFile)
+	http.ServeFile(w, r, cachePath)
 	return nil
 }
 ```

Duncaen:
Yes looks good to me.

And yes clients I looked at use different sizes, Jamstash uses 45 and 160, Aurial 30 and 200 and DSub doesn't set a size so it will use 600 which helps a lot with some of the covers I have.

sentriz:
ok added one more commit. just puts the scaled covers in a "covers" subdirectory of the cache directory. if you're ok with that commit i think we're good to merge

sentriz:
thanks!

Issue #68, opened 2020-05-18
Support libre.fm


gjabell:
It would be great if gonic supported [libre.fm](https://libre.fm/) as an alternative to Last.fm. Libre.fm should implement the same scrobbling API as Last.fm, so the only differences would need to be the authentication flow.

If this is something you'd consider adding to the project I'd be happy to help with the implementation!

sentriz:
hey
yeah I think this would be a nice addition. perhaps creating a "scrobble" abstraction that will try submit to all of lastfm, librefm, etc that exist

gjabell:
Great! If I have some free time this week I'll take a crack at
implementation.

Duncaen:
Would also be nice if the scrobble abstraction can also update local playcounts when submission mode is used so clients like DSub can update the playcount if they are used offline.

gjabell:
That should be pretty easy to add :)
I will be moving soon and won't have much time to work on this for a few weeks, but hopefully I can get some code up before too long

spezifisch:
> yeah I think this would be a nice addition. perhaps creating a "scrobble" abstraction that will try submit to all of lastfm, librefm, etc that exist

Any updates on this?

This would be relevant to my work in progress to add ListenBrainz scrobbling support in my repo at https://github.com/spezifisch/gonic/commit/4a6b905493d2d46eba7b56d9405743c2f8a34311 (it's working, but I need to do some cleanup before I would create an issue for this).

PR #90, opened 2020-09-16
Add ListenBrainz API scrobbling support

spezifisch:
This PR adds:
* a settings page for the admin to enable ListenBrainz support globally (disabled by default)
* optionally allow users to specify a custom API endpoint. This is useful if they're running their own ListenBrainz server. It's disabled by default (only `api.listenbrainz.org` is allowed), because users could specify any server here. But you can enable this if you trust your users.
* users can enter their ListenBrainz API token (obtained from their LB profile page) to scrobble to their account
* code for the `ServeScrobble` handler. Last.fm, ListenBrainz, or both are used if configured by the user.
* some style fixes

Todo:
* Tests could be better

Tested with:
* Servers: ListenBrainz official site, musicbanana dev version
* Clients: Ultrasonic, Strawberry dev version, DSub (from F-Droid; doesn't call scrobble.view at all?)

Related:
* a scrobbler abstraction like mentioned in #68 would be nice.

### DIFF
--- a/server/assets/pages/home.tmpl
+++ b/server/assets/pages/home.tmpl
@@ -43,7 +43,7 @@
             <p class="text-light">api key not set</p>
             {{ if not .User.IsAdmin }}
                 <p class="text-light">please ask your admin to set it</p>
-            {{ end }}
+	          {{ end }}
         {{ end }}
         {{ if .User.IsAdmin }}
             <p><a href="{{ path "/admin/update_lastfm_api_key" }}">update api key&#8230;</a></p>
@@ -69,8 +69,8 @@
             {{ else }}
                 <span class="angry">unlinked</span>
 
-                <table id="listenbrainz-pref">
                 <form id="listenbrainz-pref-set" action="{{ path "/admin/link_listenbrainz_do" }}" method="post"></form>
+                <table id="listenbrainz-pref">
                 <tr>
                     <td><label for="listenbrainz-token">Token:</label></td>
                     <td><input form="listenbrainz-pref-set" id="listenbrainz-token" type="text" name="token" value="{{ default "" .User.ListenBrainzToken }}"></td>
--- a/server/assets/pages/home.tmpl
+++ b/server/assets/pages/home.tmpl
@@ -43,10 +43,57 @@
             <p class="text-light">api key not set</p>
             {{ if not .User.IsAdmin }}
                 <p class="text-light">please ask your admin to set it</p>
-	    {{ end }}
+            {{ end }}
         {{ end }}
         {{ if .User.IsAdmin }}
             <p><a href="{{ path "/admin/update_lastfm_api_key" }}">update api key&#8230;</a></p>
+        {{ end }}
+    </div>
+</div>
+<div class="padded box">
+    <div class="box-title">
+        <i class="mdi mdi-brain"></i> ListenBrainz
+    </div>
+    <div class="box-description text-light">
+        <p>gonic can scrobble to <a href="https://listenbrainz.org/" target="_blank">ListenBrainz</a> and compatible sites.</p>
+    </div>
+    <div class="text-right">
+        {{ if .ListenBrainzEnabled }}
+            <span class="text-light">current status</span>
+            {{ if .User.ListenBrainzToken }}
+                linked to <code>{{ .User.ListenBrainzURL  }}</code>
+                <span class="text-light">&#124;</span>
+                <form action="{{ path "/admin/unlink_listenbrainz_do" }}" method="post">
+                    <input type="submit" value="unlink">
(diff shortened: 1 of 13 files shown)


sentriz:
very cool! thanks very much for the new feature. i will try my best to review it this weekend :)

Duncaen:
> DSub (from F-Droid; doesn't call scrobble.view at all?)

You you go offline in the sidebar, play some tracks and go online again it will ask if it should scrobble and then provides a list of tracks/timestamps with the "submission" parameter set.

sentriz:
> DSub will use search3 to find the trackid with unsupported queries.
I've seen this too, it seems at least in my case it that sometimes calls scrobble.view with the track ID, then sometimes it uses some sort of psuedo SQL query. do you know that the criteria is for track ID vs query?

sentriz:
hey sorry about leaving this for so long @spezifisch 
In general I like it, thanks! currently I'm wondering about the need for the "Allow users to use ListenBrain" and " Allow users to set a custom API endpoint " options

are there any drawbacks to having these both enabled by default (and then no need for a new admin view)
everyone could be presented with this screen by default

image

with `https://api.listenbrainz.org` prefilled by default, but possible to change

Duncaen:
> I've seen this too, it seems at least in my case it that sometimes calls scrobble.view with the track ID, then sometimes it uses some sort of psuedo SQL query. do you know that the criteria is for track ID vs query?

This is weird, looking further into it I can't also see that it sometimes works and sometimes not. Looking at the DSub code, it should get the song ID based on its cached path, but it uses the ID as path, maybe in some ways scrobble is called with the filepath as ID where it will successfully scrobble and then otherwise it will use the search criteria.
 
https://github.com/daneren2005/Subsonic/blob/01e15feec61d8fea4d424d002a6a9cde1c23a82e/app/src/main/java/github/daneren2005/dsub/service/OfflineMusicService.java#L561

spezifisch:
Thanks for your feedback! 

> are there any drawbacks to having these both enabled by default (and then no need for a new admin view)
> everyone could be presented with this screen by default

I don't see any drawbacks to having Listenbrainz support enabled by default. 

Assuming that all your users are good, there should also be no drawback to allowing custom URLs by default. Though I see a security consideration in that you're basically allowing users to have the server make HTTP POST requests to arbitrary URLs albeit with a pretty limited HTTP body. Considering also that there is no rate limiting on calling `scrobble.view`, I could imagine that an evil user might exploit this somehow as a proxy.

sentriz:
> you're basically allowing users to have the server make HTTP POST requests to arbitrary URLs 

yeah I had considered this too, but personally the people who use my server are strictly friends and family. i would (imagine) it's the same for most people? I can't imagine there are "public" gonic servers

in this case i think it's a small potential security risk vs. much simpler code. at the moment I prefer the latter :D what do you think?

sentriz:
> > I've seen this too, it seems at least in my case it that sometimes calls scrobble.view with the track ID, then sometimes it uses some sort of psuedo SQL query. do you know that the criteria is for track ID vs query?
> 
> This is weird, looking further into it I also see that it sometimes works and sometimes not. Looking at the DSub code, it should get the song ID based on its cached path, but it uses the ID as path, maybe in some ways scrobble is called with the filepath as ID where it will successfully scrobble and then otherwise it will use the search criteria.
> 
> https://github.com/daneren2005/Subsonic/blob/01e15feec61d8fea4d424d002a6a9cde1c23a82e/app/src/main/java/github/daneren2005/dsub/service/OfflineMusicService.java#L561

that is very strange, maybe I can ask daneren2005, he seems to be be more active on the repo these days :) I don't have a setup for running a debug DSub build either

spezifisch:
> > you're basically allowing users to have the server make HTTP POST requests to arbitrary URLs
> 
> yeah I had considered this too, but personally the people who use my server are strictly friends and family. i would (imagine) it's the same for most people? I can't imagine there are "public" gonic servers
> 
> in this case i think it's a small potential security risk vs. much simpler code. at the moment I prefer the latter :D what do you think?

Good point. I'm in favour of removing the new admin view to keep it simple but maybe somewhere the admin should be made aware of our assumption that users are trustworthy. :smiley:

Duncaen:
This PR needs to enable scrobbling in ServeGetUser, thats why DSub doesn't scrobble in Online mode with only a listenbrainz session at least.

sentriz:
hey @spezifisch thank you for your work on this
since #107 popped up, i thought it slightly better for the cool Scrobbler interface (and it still has your ui components) 
it also takes a simpler approach to permissions

hope this is ok, and thanks again

sentriz:
also apologies it's taken so long to do this

PR #93, opened 2020-09-20
Revamp the transcoding decision tree

fijam:
I made changes to the transcoding decisions to accommodate client requests. This is to avoid transcoding lossy file formats like mp3s when not necessary.

The logging is pretty verbose to show what is going on, feel free to remove it.

**Example 1**. The client  requests bitrate less than 320k, requested file is FLAC
```
2020/09/20 11:00:32 client requests 320k, transcoding profile 128k, track bitrate 572k
2020/09/20 11:00:32 Transcoding according to client request of 320k
2020/09/20 11:00:32 serving transcode `03 Lucky To Be Me.flac`: cache [mp3/320k] miss!
```
The client preference (320k) takes precedence over the server profile (128k) setting. This effectively dynamically adapts the transcoding profile to what the client wants.

**Example 2**. The client  requests bitrate less than 320k, requested file is mp3 
```
2020/09/20 10:56:27 client requests 320k, transcoding profile 128k, track bitrate 240k
2020/09/20 10:56:27 Not transcoding, requested bitrate larger or equal to track bitrate
2020/09/20 10:56:27 serving raw "04 - Detlef Schrempf.mp3"
```
Even though the file's bitrate is larger than the transcoding profile, it still matches client preferences so it's served raw to avoid transcoding. 

**Example 3** The client has no preference (maxBitRate=0), requested file is FLAC
```
2020/09/20 11:05:37 client requests 0k, transcoding profile 128k, track bitrate 722k
2020/09/20 11:05:37 Transcoding according to transcoding profile of 128k
2020/09/20 11:05:37 serving transcode `08 Don´t Wait.flac`: cache [mp3/128k] miss!
````
If the client requests maxBitRate=0 (unlimited), the server transcoding profile applies.

**Example 4** The client has no preference (maxBitRate=0), requested file is a small mp3
```
2020/09/20 11:08:58 client requests 0k, transcoding profile 128k, track bitrate 128k
2020/09/20 11:08:58 Not transcoding, transcoding profile bitrate larger or equal to track bitrate
2020/09/20 11:08:58 serving raw "Jethro Tull - 07 - Too Old To Rock 'n' Roll, Too Young To Die.mp3"
```

Note that with the current behavior, all four examples would be transcoded to 128k.

This is a fairly simple change, but also my first time dealing with Go so please review for mistakes.

### DIFF
--- a/server/ctrlsubsonic/handlers_raw.go
+++ b/server/ctrlsubsonic/handlers_raw.go
@@ -168,6 +168,7 @@ func (c *Controller) ServeStream(w http.ResponseWriter, r *http.Request) *spec.R
 	}
 	encodeOptions := encode.Options{
 		TrackPath:        trackPath,
+		TrackBitrate:     track.Bitrate,
 		CachePath:        c.CachePath,
 		ProfileName:      pref.Profile,
 		PreferredBitrate: params.GetOrInt("maxBitRate", 0),
--- a/server/encode/encode.go
+++ b/server/encode/encode.go
@@ -146,14 +146,6 @@ func cacheKey(sourcePath string, profileName string, profile Profile) string {
 	)
 }
 
-// getBitrate checks if the client forces bitrate lower than set in profile
-func getBitrate(preferred, defined int) int {
-	if preferred != 0 && preferred < defined {
-		return preferred
-	}
-	return defined
-}
-
 type (
 	OnInvalidProfileFunc func() error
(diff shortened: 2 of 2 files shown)


sentriz:
thanks! I added some small personal style preferences. also this "Encode" abstraction I made is not good / kinda weird I think, so this may be moved around in future. but will keep these changes

Issue #31, opened 2020-01-30
Ratings support?


sardaukar:
Is implementing ratings per track a goal?

sentriz:
yes I would hope to do this in the future. should not be too hard. just need some time

davidhampgonsalves:
@sentriz do you have ideas for how you plan on implementing this? I see that nicksellen/audiotags only supports read operations through taglib. Were you thinking of adding the ability to change ratings to that first?

sentriz:
You mean to say the ratings are stored in the audio tags themselves?

sentriz:
My idea was a track rating table instead of tags. With columns track_id, user_id, rating - to make sure ratings are per user. But I'd like to know what airsonic etc does

davidhampgonsalves:
Looks like that's what airsonic is doing as well:

https://github.com/airsonic/airsonic/blob/828c1b2674b1b5bf4f96454ed5c154a0c074f3bb/airsonic-main/src/main/java/org/airsonic/player/controller/SetRatingController.java

I need the files to be the source of truth so I'll do something different. 
Thanks

sentriz:
if you follow that path:
https://github.com/airsonic/airsonic/blob/828c1b2674b1b5bf4f96454ed5c154a0c074f3bb/airsonic-main/src/main/java/org/airsonic/player/controller/SetRatingController.java#L62
-> 
https://github.com/airsonic/airsonic/blob/828c1b2674b1b5bf4f96454ed5c154a0c074f3bb/airsonic-main/src/main/java/org/airsonic/player/service/RatingService.java#L76
->
https://github.com/airsonic/airsonic/blob/828c1b2674b1b5bf4f96454ed5c154a0c074f3bb/airsonic-main/src/main/java/org/airsonic/player/dao/RatingDao.java#L80

they seem to be doing the same as I suggested

davidhampgonsalves:
Yes, sorry I was unclear, I meant that what you suggested is what airsonic does.

sentriz:
ohh sorry yes. I misread :) if I did end up implementing ratings the way airsonic does. you could probably make a little script to update the gonic db's ratings for a given user and list of tracks

davidhampgonsalves:
Yea I think I would do that. Also thanks for exposing gonicscan, really handy.

sentriz:
also, looking at http://www.subsonic.org/pages/api.jsp "setRating"
this depends on #41 
[...]

PR #95, opened 2020-10-25
support Track setRating 

davidhampgonsalves:
This adds support for setRating (but only for track). If this looks acceptable I will make similar changes for Artist / Album and maybe stars.

https://github.com/sentriz/gonic/issues/31


sentriz:
cool thanks! though one issue with storing the rating directly in the track table is that ratings aren't per-user. if one user rates a track some value, all users have rated that track in a way.
what might be better is a many to many table with columns, for example, `user_id`, `track_id`, `rating`. similar to the `transcode_preference` table

davidhampgonsalves:
I use this as a single user and forgot all about the others. I'll make that change.

davidhampgonsalves:
Took a cursory look at this and it looks like for linking the rating to the artist, track, album I will either need: 
 - separate tables for track, album, artist ratings
 - single table with attributes for each type (track_id, artist_id, album_id)

I took a quick look at airsonic and it uses the media path rather then an id to create the linkage but I don't think that's inline with gonic.

Do you have opinions on how you would like to see this done (I'd probably go with a single table for simplicity with index on each id type for query performance).

PR #96, opened 2020-11-06
Allow admin to set encoding profiles for all users

abyxcos:
I am not sure if this is the intended usage, but I only log in to Gonic as the admin user, and I expected encoding settings to be global. This allows all users to pick up the admin settings so you don't have to re-do these each time.

### DIFF
--- a/server/ctrlsubsonic/handlers_raw.go
+++ b/server/ctrlsubsonic/handlers_raw.go
@@ -28,7 +28,7 @@ import (
 func streamGetTransPref(dbc *db.DB, userID int, client string) db.TranscodePreference {
 	pref := db.TranscodePreference{}
 	dbc.
-		Where("user_id=?", userID).
+		Where("user_id=1 OR user_id=?", userID).
 		Where("client COLLATE NOCASE IN (?)", []string{"*", client}).
 		Order("client DESC"). // ensure "*" is last if it's there
 		First(&pref)


sentriz:
the idea with this is that all users choose their own trancoding perferences. personally I like it this way. also with this change, if I understand it right, means that if the admin has transcoding enabled for just one client, say DSub - and a normal user has no preferences - then they now how transcoding enabled just for DSub too (seems strange to me)

abyxcos:
This patch makes some assumptions about usage. My admin user is only used for the sake of forcing scans, and I don’t have any browsers that can play flac (nor do any of my users). If you actively use admin as a user, then that breaks this model. 

This could be approached some other ways, like allowing transcoding rules to be enabled/disabled and then listing the admin ones as “defaults” for all users, but that’s much more invasive, and possibly a privacy leak depending on your usage.

nstetter:
I second having a global override for transcoding. So in the admin account I could set a fallback for any client to use mp3 transcoding so gonic never serves flac or other formats, thus saving bandwith.

sentriz:
hope you don't mind, I'm gonna close this for now. I think in a way it's slightly unexpected behaviour to fallback to a specific users transcoding settings (even if the user is admin)

maybe we could open a new issue to explicitly set global fallback options elsewhere in the UI

PR #97, opened 2020-11-12
Wrap the filePath in quotes

abyxcos:
Encoding a filename with spaces (or possibly dashes?) seems to get messy, so best to just wrap it in quotes so ffmpeg doesn't do anything smart with it.


abyxcos:
I take this back, it looks like my problem was around symlinking ffmpeg into /usr/bin/ffmpeg (path on my platform is /usr/local/bin/ffmpeg).

PR #99, opened 2020-11-25
Mark cache files as partial until `ffmpeg` finishes without errors

spijet:
At the moment Gonic writes data that comes from `ffmpeg` to cache files right away, and it may result in incomplete cache files (e.g. if either `ffmpeg` or Gonic crash mid-transcode). This PR adds a fairly stupid workaround: write to a `.part` file and rename it after it's been written to successfully.


sentriz:
makes sense thanks!

PR #102, opened 2020-12-12
Add stmp to supported clients

wildeyedskies:
stmp is a terminal based subsonic client. I've developed it specifically for use with gonic.


sentriz:
cool thanks!

PR #104, opened 2020-12-16
multiple genre support and other stuff

sentriz:


sentriz:
hey @Duncaen 

this PR adds support for multiple genres. eg. in my files I have tags such as `drum and bass;acid;electronic` where it would be good to show search those tag individually 

what do you think? anything you would change? (since you worked on this stuff initially :D )

keep in mind 
 - there should (hopefully) be nothing changed from a users perspective because of the migration from one to many -> many to many
 - the scanner genre db stuff is not good (inserting records in a loop) but i would like to upgrade to gorm v2 in the future which would provide batch insert etc
 - i still need to test this some more myself

Duncaen:
Seems to work fine, only issue I see is `UNIQUE constraint failed: track_genres.track_id, track_genres.genre_id` for every track when doing a full re-scan.

sentriz:
> Seems to work fine, only issue I see is `UNIQUE constraint failed: track_genres.track_id, track_genres.genre_id` for every track when doing a full re-scan.

thanks! should be fixed now

PR #105, opened 2020-12-21
Add year to TrackChilds of an Album

gsol10:
Hello,

This commit adds the album TagYear to the year field of every TrackChilds (when using tag browsing).

This allows implementations that don't support the Year field of albums (introduced in Subsonic 1.10.1) to derive the album year anyways. For information [Ultrasonic](https://github.com/ultrasonic/ultrasonic) behaves like this. 

In my setup I only use the tag browsing, but I think it might be useful to add it to NewTCTrackByFolder:construct_by_folder:39 as well. Tell me if you want me to add it to the PR.

Thanks with your work on this project otherwise, it's very nice !


sentriz:
thanks!

PR #107, opened 2021-01-08
Add listen Listenbrainz support

lxea:
I started this before i noticed #90, however it now uses the html part of #90. I submitted this as it takes a different approach for the scrobbling, it adds a scrobbler interface, this should hopefully in the future make it easier add support for #68 

It doesn't include support for custom listenbrainz urls just for simplicity sake.

This is only fairly lightly tested, but scrobbling to both listenbrainz and last.fm still seems to work.


sentriz:
cool thank you alex!

PR #109, opened 2021-01-17
Jukebox Improvments: Support seeking, fix the default returned return api type

lxea:
Currently beep doenst support seeking with flac but it does with mp3 and (presumably wav) so it doesnt work but once https://github.com/mewkiz/flac/pull/43 merges it should be easy to add support for it to beep

### DIFF
--- a/server/jukebox/jukebox.go
+++ b/server/jukebox/jukebox.go
@@ -59,13 +59,15 @@ const (
 )
 
 type update struct {
-	action updateType
-	index  int
-	tracks []*db.Track
+	action     updateType
+	index      int
+	tracks     []*db.Track
+	skipOffset int
 }
 
 type updateSpeaker struct {
-	index int
+	index  int
+	offset int
 }
 
 func New(musicPath string) *Jukebox {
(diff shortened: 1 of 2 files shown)


server/jukebox/jukebox.go
```diff
@@ -182,14 +184,20 @@ func (j *Jukebox) doUpdateSpeaker(su updateSpeaker) error {
 	j.Lock()
 	j.info = &strmInfo{}
 	j.info.strm = streamer.(beep.StreamSeekCloser)
+	if su.offset != 0 {
+		err = j.info.strm.Seek(format.SampleRate.N(time.Second * time.Duration(su.offset)))
+		if err != nil {
+			return err
+		}
```
sentriz, line 191:
```suggestion
		samples := format.SampleRate.N(time.Second * time.Duration(su.offset))
		if err := j.info.strm.Seek(samples); err != nil {
			return fmt.Errorf("seek stream: %w", err)
		}
```

how about an intermediary var what for doing the `if err := <>; err != nil {`

sentriz:
looks good to me, though i don't have the context for the samples stuff. how does `sample rate * time` work for VBR tracks?

lxea:
> looks good to me, though i don't have the context for the samples stuff. how does `sample rate * time` work for VBR tracks?

I'm not super sure really, when i was testing it with my mp3s which were VBR it seemed to be working right, i think the .N() accounts for the sample times, its the same way the track progress works too

sentriz:
thanks mick

PR #115, opened 2021-02-04
feature: add tini to dockerfile

onedr0p:
tini is a neat little `init`

https://github.com/krallin/tini

Benefits? See here

https://github.com/krallin/tini/issues/8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,8 @@ FROM alpine:3.12.3
 RUN apk add -U --no-cache \
   ffmpeg \
   ca-certificates \
-  tzdata
+  tzdata \
+  tini
 COPY --from=builder \
   /usr/lib/libgcc_s.so.1 \
   /usr/lib/libstdc++.so.6 \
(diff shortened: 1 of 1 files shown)


zesoft23:
Random drop-in, but is it more worthwhile to add `tini` as a default argument in the README? `tini` has been built in since [Docker 19](https://github.com/krallin/tini) by adding the `--init` flag to `docker run`. It [can be used](https://stackoverflow.com/questions/50356032/whats-the-docker-compose-equivalent-of-docker-run-init#50357065) in `docker-compose` too. Seems more helpful to place the dependency on docker vs. integrating into the project. Just my 2 cents!

onedr0p:
This isn't the case in Kubernetes, where we can use cri-o or containerd not just the Docker CRI.

zesoft23:
Ah awesome, I had no idea it wasn't across the whole Docker CRI. Good luck with the PR!

sentriz:
makes sense, thanks!

Issue #117, opened 2021-02-05
 Podcast with Gonic


uGeek:
Great new functionality.

I had several doubts.

- Once subscribed to the podcasts, how is it updated?
- I have subscribed to 4 podcasts and it works perfect, but I have subscribed to mine and the episodes do not appear.
- Lastly, as a suggestion, it would be great to add an option to import podcast subscriptions from an .opml file

Thank you very much. Great job

My Feed: http://feeds.feedburner.com/uGeek
Yesterday I updated the docker ARM to the latest version

lxea:
Hey uGeek, 

>     * Once subscribed to the podcasts, how is it updated?

gonic will automatically refresh all podcasts once every hour but it wont automatically download them.

>     * I have subscribed to 4 podcasts and it works perfect, but I have subscribed to mine and the episodes do not appear.

Sorry to hear that, I'll investigate and see if i can make a fix :) 

>     * Lastly, as a suggestion, it would be great to add an option to import podcast subscriptions from an .opml file

Good idea, I'll also look into this!

uGeek:
I have done a new installation and I have not finished scanning all the music, to test the podcasts. I do not know if it has to do because they are not updated.

It could also be because I am using the docker that I have built, I will try the one from amd64.

DSub has an option to force the server to update, it seems that it works from the app, but it does not update me

lxea:
Whats not updating for you? Are new episodes not appearing? 
#118 should fix the issue of episodes from your feed here not showing up http://feeds.feedburner.com/uGeek

There isnt a force scan option for podcasts except to use the one in dsub/timer

uGeek:
I have mounted the sentriz/gonic docker for amd64, from the previous comment and it takes more than 2 hours and it does not update me. 
I have subscribed to a Radio podcast. In AntennaPod and new episodes arrive and not in the Gonic server

lxea:
Oh weird, do you mind sharing the url with me, i'll try test it and see if i can fix it

uGeek:
For example: http://api.rtve.es/api/programas/1750/audios.rss

lxea:
Hey, I've updated #118 it should hopefully fix updating for this feed.

You'll need to remove the feeds that werent updating and add them again though I think

uGeek:
Brilliant!!! 

Now my Feed works and updates all podcasts perfectly.

Good job and thank you very much

Code-Slave:
Heres a feed that doesnt work as it has a \ in the title

[...]

PR #118, opened 2021-02-05
make finding podcast feed audio more robust and use user provided feed urls

lxea:
this will check that if there is no type field in an enclosure that the url ends with an audio filename, if none of these are true, itll then check audio:content for an item and do those checks again.

this was reported to be an issue as episodes werent showing up in #117 which it should fix 

This will also fix feeds that dont contain a feed url not updating  also mentioned in #117

### DIFF
--- a/server/podcasts/podcasts.go
+++ b/server/podcasts/podcasts.go
@@ -20,6 +20,7 @@ import (
 
 	"go.senan.xyz/gonic/multierr"
 	"go.senan.xyz/gonic/server/db"
+	gmime "go.senan.xyz/gonic/server/mime"
 	"go.senan.xyz/gonic/server/scanner/tags"
 )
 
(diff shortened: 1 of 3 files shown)


server/podcasts/podcasts.go
```diff
@@ -170,26 +171,73 @@ func (p *Podcasts) AddEpisode(podcastID int, item *gofeed.Item) error {
...
+func isAudio(mediaType, url string) bool {
+	if mediaType != "" && strings.HasPrefix(mediaType, "audio") {
+		return true
+	}
+	switch filepath.Ext(url) {
+	case ".flac", ".mp3", ".ogg", ".opus", ".wav", ".alac":
+		return true
+	}
+	return false
+}
```
sentriz, line 200: do you think it could be possible to leverage `server/mime.FromExtension` here? it can return a bool
https://github.com/sentriz/gonic/blob/master/server/mime/mime.go
feel free to update it too if it could work

lxea: :+1:

sentriz:
thanks!

PR #119, opened 2021-02-06
List /podcasts as a volume in dockerfiles

edwargix:
When I pulled the latest `sentriz/gonic` from Docker Hub, my container failed to start because I didn't have anything mounted to `/podcasts`.  If we list `/podcasts` as a volume in the Dockerfile, docker should automatically create the directory if it doesn't exist.

I tested this change locally.


sentriz:
thanks!

PR #122, opened 2021-02-18
Fix podcasts with / in the name/episode title

lxea:
https://github.com/sentriz/gonic/issues/117#issuecomment-780971049

[failure is due to linting timing out, but i get nothing locally]


sentriz:
thanks!!!!!!!!!!!

PR #123, opened 2021-02-19
Download new podcast episodes automatically & Add download all button

lxea:
This makes it so that any newly released episodes will be automatically downloaded if you select the option in the interface

It also adds a button to the interfaces to download all the available episodes for a podcast
image


server/podcasts/podcasts.go
```diff
@@ -275,14 +308,38 @@ func (p *Podcasts) refreshPodcasts(podcasts []*db.Podcast) error {
...
+	if err != nil {
+		return fmt.Errorf("get episodes by podcast id: %w", err)
+	}
+	go func() {
+		for _, episode := range podcastEpisodes {
+			if episode.Status == episodeDownloading || episode.Status == episodeCompleted {
+				log.Println("skipping episode is in progress or already downloaded")
+				continue
+			}
+			if err := p.DownloadEpisode(episode.ID); err != nil {
```
lxea, line 334: Im not sure if we want a sleep(5) or something here? I didnt notice issues with it but maybe some hosting sites would rate limit ¯\\_(ツ)_/¯

sentriz:
thanks ali!!!

PR #132, opened 2021-04-18
scanner/tags: prefer original_date over date

Duncaen:
I have some re-released albums tagged with musicbrainz which sets the `date`/`year` to the date of the re-release and adds `original_date`/`original_year` to the initial release date.

I personally prefer to see the original dates when browsing my music, not sure if there is a case against it.


sentriz:
Makes sense thanks!

PR #140, opened 2021-05-09
ci: use short hash

jef:
### Description

You mentioned how you thought about using a short hash instead of the longer one. This would do the trick. Although seems ugly, may be one of the work arounds. You could combine with another step if you'd like as well.


sentriz:
thanks!

Issue #63, opened 2020-04-29
Incremental scan does not add new tracks


Flameborn:
I currently have a Gonic database with 83007 tracks. After Gonic finishes running an incremental scan, I get:

```
Apr 29 14:00:22 gonic[318]: 2020/04/29 14:00:22 finished scan in 4m33.9894
92757s, +727/83734 tracks (19 err)
```

The errors are related to tag reading from a few flac and m4a files.

After this, I see the new number of tracks as 83734 via the web interface, which is correct. Then, cleanup runs, and I get:

```
Apr 29 14:02:53 idris gonic[318]: 2020/04/29 14:02:53 finished clean in 2m30.910
381764s, -0 tracks
```

Right after this, the number of tracks reverts to 83007 tracks again.

What is interesting is that even if I run a full scan, Gonic does not seem to pick up all the tracks.

Maximalfr:
I have the same issue and I think mp3 files are impacted too

sentriz:
ok having a look now. it seems for some reason the cleaning logic is deleting tracks it already added just before?

sentriz:
good news, I can reproduce it

Flameborn:
Awesome! Is it related to the cleaning step?

sentriz:
hi! can you try `sentriz/gonic:clean-refac` please

Flameborn:
Hello,

This time, I got one less than I had before the incremental scan started.

sentriz:
The branch is https://github.com/sentriz/gonic/tree/scan-clean-del

Flameborn:
Ah, thanks.

It seems that we have the same issue as before. Starting at *83004* via an incremental scan:

```
May 01 18:00:21 idris gonic[21949]: 2020/05/01 18:00:21 finished scan in 15m13.81595s, +730/83735 tracks (19 err)
May 01 18:00:33 idris gonic[21949]: 2020/05/01 18:00:33 finished clean tracks in 12.42918s, 0 removed
May 01 18:00:33 idris gonic[21949]: 2020/05/01 18:00:33 finished clean folders in 482.32ms, 0 removed
May 01 18:02:12 idris gonic[21949]: 2020/05/01 18:02:12 finished clean artists in 1m38.54148s, 403 removed
```

sentriz:
hmm that is strange indeed. also can't understand why your times are so slow. you only have less than 3x the tracks I have but mine are like

```
[...]

PR #145, opened 2021-06-20
Strip EBU R128 gain tags when using forced-RG transcoding

spijet:
Hello @sentriz, long time no see! :)

I was messing around with my encode profiles and found that in some cases (e.g. my music library) source files can have EBU R128 gain tags alongside ordinary ReplayGain tags. If we don't instruct `ffmpeg` to strip out these tags during RG-enabled transcoding, it can confuse some client apps into thinking that the track has ReplayGain tags and should be processed accordingly. 

The main problem is that these R128 tags have a different scale than normal RG ones, and even when left with a value of `0` (zero) it can cause some clients to forcibly apply a whopping +5.0 dB gain to the track, causing clipping and occasional ear bleeding. :)

This PR fixes it by stripping these tags out. I double-checked the result (with `mediainfo` and Foobar2000) and now there are no rogue gain tags left anymore.

### DIFF
--- a/server/encode/encode.go
+++ b/server/encode/encode.go
@@ -102,6 +102,8 @@ func ffmpegCommand(filePath string, profile Profile) (*exec.Cmd, error) {
 			"-metadata", "replaygain_album_peak=",
 			"-metadata", "replaygain_track_gain=",
 			"-metadata", "replaygain_track_peak=",
+			"-metadata", "r128_album_gain=",
+			"-metadata", "r128_track_gain=",
 		)
 	}
 	args = append(args, "-f", profile.Format, "-")


sentriz:
hi again! thanks for the fix :) this makes sense to me. the next release will hopefully be coming after #63 is fixed

spijet:
Wow, that was quick! :D

I'll add more PRs tomorrow — my local branch now has a bit more accurate (but tad slower) RG processing and some other quality-of-life additions, like properly specifying the MIME-type when transcoding (on cache hit AND cache miss) and `ffmpeg` now sending data to `gonic` in chunks of 16K bytes (this kiiinda helps with TTFB being small). As for now, it's 2 AM in my timezone. :)

sentriz:
cool! looking forward to see what you have :+1: good night

PR #146, opened 2021-06-21
Add MIME-type and content length hints to cache handlers

spijet:
This PR adds a simple MIME-type guessing logic to `onCacheHit` and `onCacheMiss` handlers, which sets `Content-Type` HTTP response header based on format specified by transcoding profile. It also sets `Content-Length` header, but only on cache hit (as we can't reliably know response length on cache miss, because VBR is a thing).

### DIFF
--- a/server/ctrlsubsonic/handlers_raw.go
+++ b/server/ctrlsubsonic/handlers_raw.go
@@ -235,13 +235,12 @@ func (c *Controller) ServeStream(w http.ResponseWriter, r *http.Request) *spec.R
 		cacheMime, _ := mime.FromExtension(profile.Format)
 		w.Header().Set("Content-Type", cacheMime)
 
-		cacheFile, cfErr := os.Stat(path)
-		if cfErr != nil {
-			log.Printf("failed to stat cache file `%s`: %v", path, cfErr)
-		} else {
-			contentLength := fmt.Sprintf("%d", cacheFile.Size())
-			w.Header().Set("Content-Length", contentLength)
+		cacheFile, err := os.Stat(path)
+		if err != nil {
+			return fmt.Errorf("failed to stat cache file `%s`: %v", path, err)
 		}
+		contentLength := fmt.Sprintf("%d", cacheFile.Size())
+		w.Header().Set("Content-Length", contentLength)
 		http.ServeFile(w, r, path)
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


server/ctrlsubsonic/handlers_raw.go
```diff
@@ -231,12 +232,24 @@ func (c *Controller) ServeStream(w http.ResponseWriter, r *http.Request) *spec.R
...
+		cacheMime, _ := mime.FromExtension(profile.Format)
+		w.Header().Set("Content-Type", cacheMime)
+
+		cacheFile, cfErr := os.Stat(path)
+		if cfErr != nil {
+			log.Printf("failed to stat cache file `%s`: %v", path, cfErr)
+		} else {
+			contentLength := fmt.Sprintf("%d", cacheFile.Size())
+			w.Header().Set("Content-Length", contentLength)
+		}
```
sentriz, line 244: i think it's safe to return a real error here because if the stat fails, then the http.ServeFile will fail. thoughts?

```suggestion
		cacheFile, err := os.Stat(path)
		if err != nil {
			return fmt.Errorf("failed to stat cache file `%s`: %v", path, err)
		}
		contentLength := fmt.Sprintf("%d", cacheFile.Size())
		w.Header().Set("Content-Length", contentLength)

```

sentriz (changes requested):

server/ctrlsubsonic/handlers_raw.go
spijet: Sounds reasonable, yes. :)

server/ctrlsubsonic/handlers_raw.go
```diff
@@ -231,12 +232,23 @@ func (c *Controller) ServeStream(w http.ResponseWriter, r *http.Request) *spec.R
 	onCacheHit := func(profile encode.Profile, path string) error {
 		log.Printf("serving transcode `%s`: cache [%s/%dk] hit!\n",
 			audioFile.AudioFilename(), profile.Format, profile.Bitrate)
+		cacheMime, _ := mime.FromExtension(profile.Format)
+		w.Header().Set("Content-Type", cacheMime)
+
+		cacheFile, err := os.Stat(path)
+		if err != nil {
+			return fmt.Errorf("failed to stat cache file `%s`: %v", path, err)
```
sentriz, line 240:
```suggestion
			return fmt.Errorf("failed to stat cache file `%s`: %w", path, err)
```

sentriz: cheeky linter

sentriz:
cool thanks! :)

Issue #121, opened 2021-02-17
Adjust transcoding bitrate


amitjethani:
Love the ability to set a transcode profile based on client. However, I request a further enhancement where the server admin can specify the bitrate as well as 128 kbps is on the lower side of what I would prefer. Ideally this is a configuration variable that can be specified as a Docker environment variable or can be adjusted directly on the management page.

sentriz:
thats a cool idea. the meantime though, maybe it could be nice to just provide more presets, other than the 4 we have now? what sort of bitrates etc would you want to see?

a10software:
That would work just as well! I'm drawing some inspiration from a popular music app's settings panel, but here is one idea for presets: 
- 64 kbps (for bandwidth constrained/audiobooks)
- 160 kbps
- 320 kbps
- lossless (no transcoding)

spijet:
Unfortunately, Subsonic API doesn't provide a way to "say" the server that you want to stream something without transcoding, IIRC. It does provide a `/download` endpoint though (which will give you the requested track exactly as it's stored on the server), but I'm not sure if it's supposed to send the data the same way as `/stream` in terms of suitability for (near) real-time streaming.

As for custom bitrates: current transcoding implementation allows you (the client) to specify your desired bitrate. See the this **iSub** screenshot for example.
![image](https://user-images.githubusercontent.com/14221126/123753479-7b73cc00-d8ec-11eb-88f5-8d9c181a171f.png)

If your client app sends the desired bitrate as a request parameter, and if it's lower than source track's bitrate, it'll be transcoded to the bitrate your app requested:
```golang
	case opts.PreferredBitrate != 0 && opts.PreferredBitrate < opts.TrackBitrate:
		profile.Bitrate = opts.PreferredBitrate
		log.Printf("transcoding according to client request of %dk \n", profile.Bitrate)
```

scrufulufugus:
Airsonic (and I would assume Subsonic) seems to support clients specifying `format=raw` to get and un-transcoded stream. The [play:Sub](https://apps.apple.com/us/app/play-sub-music-streamer/id955329386) client uses it if you specify and unlimited bitrate.

sentriz:
gonna close in favour of #244

PR #147, opened 2021-06-21
feat(encode): rework encoding profiles to use true VBR and add optional upsample support

spijet:
Opus encoder currently uses contstrained VBR, which seems to limit its efficiency in most cases. This PR switches Opus profiles to unconstrained VBR and also adds few tiny additions (see in-code comments for details).

### DIFF
--- a/server/encode/encode.go
+++ b/server/encode/encode.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"strings"
 
 	"github.com/cespare/xxhash"
 )
@@ -24,6 +25,8 @@ type Profile struct {
 	Bitrate       int
 	ffmpegOptions []string
 	forceRG       bool
+	hiGainRG      bool
+	upsample      bool
 }
 
 func fileExists(filename string) bool {
@@ -36,10 +39,11 @@ func fileExists(filename string) bool {
 
 func Profiles() map[string]Profile {
 	return map[string]Profile{
-		"mp3":     {"mp3", 128, []string{"-c:a", "libmp3lame"}, false},
-		"mp3_rg":  {"mp3", 128, []string{"-c:a", "libmp3lame"}, true},
-		"opus":    {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, false},
-		"opus_rg": {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, true},
+		"mp3":      {"mp3", 128, []string{"-c:a", "libmp3lame"}, false, false, false},
+		"mp3_rg":   {"mp3", 128, []string{"-c:a", "libmp3lame"}, true, false, false},
+		"opus":     {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, false, false, false},
+		"opus_rg":  {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, true, false, false},
+		"opus_car": {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, true, true, true},
 	}
 }
 
@@ -94,9 +98,33 @@ func ffmpegCommand(filePath string, profile Profile) (*exec.Cmd, error) {
 	}
 	args = append(args, profile.ffmpegOptions...)
 	if profile.forceRG {
+		aBaselineGain := 6
+		if profile.hiGainRG {
+			// This baseline gain results in final track being +3~5dB louder
+			// than Foobar2000's default ReplayGain target volume.
+			// This makes it easier to listen to music in a car, where all other
+			// sources are usually ten thousand times louder than RG-adjusted music.
+			// -- @spijet
+			aBaselineGain = 15
+		}
+		aFilters := []string{
+			fmt.Sprintf("volume=replaygain=track:replaygain_preamp=%ddB:replaygain_noclip=0", aBaselineGain),
+			"alimiter=level=disabled",
+			"asidedata=mode=delete:type=REPLAYGAIN",
+		}
+
+		// opus always forces output to 48kHz sampling rate, but we can still use upsampling
+		// to increase RG and alimiter's peak limiting precision, which is desirable in some
+		// cases. ffmpeg's `soxr` resampler is quite fast on x86-64: it takes around 5 seconds
+		// on my Ryzen 3600 to transcode an 8-minute FLAC with 2x upsample and RG applied.
+		// -- @spijet
+		if profile.upsample {
+			aFilters = append([]string{"aresample=96000:resampler=soxr"}, aFilters...)
+		}
+		aFilterString := strings.Join(aFilters, ", ")
 		args = append(args,
 			// set up replaygain processing
-			"-af", "volume=replaygain=track:replaygain_preamp=6dB:replaygain_noclip=0, alimiter=level=disabled",
+			"-af", aFilterString,
 			// drop redundant replaygain tags
 			"-metadata", "replaygain_album_gain=",
 			"-metadata", "replaygain_album_peak=",
--- a/server/encode/encode.go
+++ b/server/encode/encode.go
@@ -41,9 +41,9 @@ func Profiles() map[string]Profile {
 	return map[string]Profile{
 		"mp3":      {"mp3", 128, []string{"-c:a", "libmp3lame"}, false, false, false},
 		"mp3_rg":   {"mp3", 128, []string{"-c:a", "libmp3lame"}, true, false, false},
-		"opus":     {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, false, false, false},
-		"opus_rg":  {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, true, false, false},
-		"opus_car": {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, true, true, true},
+		"opus":     {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, false, false, false},
+		"opus_rg":  {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, true, false, false},
+		"opus_car": {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, true, true, true},
 	}
 }
 


spijet:
I just saw that this PR's branch was spun off the #146 branch and not `master`. Should I rebase it to current master instead?

sentriz:
yep I'd say this can be rebased now :+1:

spijet:
@sentriz, done! I hope I didn't break anything in the process. :)

server/encode/encode.go
```diff
@@ -94,9 +98,33 @@ func ffmpegCommand(filePath string, profile Profile) (*exec.Cmd, error) {
...
+		}
+
+		// opus always forces output to 48kHz sampling rate, but we can still use upsampling
+		// to increase RG and alimiter's peak limiting precision, which is desirable in some
+		// cases. ffmpeg's `soxr` resampler is quite fast on x86-64: it takes around 5 seconds
+		// on my Ryzen 3600 to transcode an 8-minute FLAC with 2x upsample and RG applied.
+		// -- @spijet
+		if profile.upsample {
+			aFilters = append([]string{"aresample=96000:resampler=soxr"}, aFilters...)
+		}
```
sentriz, line 123: nit, if the order doesn't matter here this can be a bit simpler
```suggestion
		if profile.upsample {
			aFilters = append(aFilters, "aresample=96000:resampler=soxr")
		}
```

server/encode/encode.go
```diff
@@ -36,10 +39,11 @@ func fileExists(filename string) bool {
...
 	return map[string]Profile{
-		"mp3":     {"mp3", 128, []string{"-c:a", "libmp3lame"}, false},
-		"mp3_rg":  {"mp3", 128, []string{"-c:a", "libmp3lame"}, true},
-		"opus":    {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, false},
-		"opus_rg": {"opus", 96, []string{"-c:a", "libopus", "-vbr", "constrained"}, true},
+		"mp3":      {"mp3", 128, []string{"-c:a", "libmp3lame"}, false, false, false},
+		"mp3_rg":   {"mp3", 128, []string{"-c:a", "libmp3lame"}, true, false, false},
+		"opus":     {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, false, false, false},
+		"opus_rg":  {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, true, false, false},
+		"opus_car": {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, true, true, true},
```
sentriz, line 46: now with all these booleans, it's little bit hard to know what's going on with for example `false, false, false`

maybe we could add some field names, and use go's false bool zero-value?

```suggestion
		"mp3": {
			Format:        "mp3",
			Bitrate:       128,
			ffmpegOptions: []string{"-c:a", "libmp3lame"},
		},
		"mp3_rg": {
			Format:        "mp3",
			Bitrate:       128,
			ffmpegOptions: []string{"-c:a", "libmp3lame"},
			forceRG:       true,
		},
		"opus": {
			Format:        "opus",
			Bitrate:       96,
			ffmpegOptions: []string{"-c:a", "libopus", "-vbr", "on"},
		},
		"opus_rg": {
			Format:        "opus",
			Bitrate:       96,
			ffmpegOptions: []string{"-c:a", "libopus", "-vbr", "on"},
			forceRG:       true,
		},
		"opus_car": {
			Format:        "opus",
			Bitrate:       96,
			ffmpegOptions: []string{"-c:a", "libopus", "-vbr", "on"},
			forceRG:       true,
			hiGainRG:      true,
			upsample:      true,
		},
```

sentriz (commented):
thanks, this is pretty cool :)

server/encode/encode.go
spijet: The order is important here (at least from what I understand from `ffmpeg`'s audio filter chain docs), as `ffmpeg` needs to resample the audio to higher rate *before* applying RG and limiting the output. It'll downsample the audio to 48kHz later (that's what Opus encoder does unconditionally, as it's part of its standard).

sentriz: Gotcha thanks 👍

server/encode/encode.go
spijet: I was thinking about converting `forceRG` to an Enum (or something similar) but never tried it.
Overall this format seems nice, but it's starting to look like it's only a matter of time before we have to move it "outside" to a YAML config file or some other place. :D

sentriz: yeah the enum is interesting. do you  think something like this could work?

```diff
diff --git a/server/encode/encode.go b/server/encode/encode.go
index 550b6f7..58d0a0b 100644
--- a/server/encode/encode.go
+++ b/server/encode/encode.go
@@ -20,13 +20,20 @@ const (
 	ffmpeg  = "ffmpeg"
 )
 
+type ReplayGain int
+
+const (
+	RGNone ReplayGain = iota
+	RGForce
+	RGHigh
+)
+
 type Profile struct {
-	Format        string
-	Bitrate       int
-	ffmpegOptions []string
-	forceRG       bool
-	hiGainRG      bool
-	upsample      bool
+	Format     string
+	Bitrate    int
+	Options    []string
+	ReplayGain ReplayGain
+	Upsample   bool
 }
 
 func fileExists(filename string) bool {
@@ -39,11 +46,35 @@ func fileExists(filename string) bool {
 
 func Profiles() map[string]Profile {
 	return map[string]Profile{
-		"mp3":      {"mp3", 128, []string{"-c:a", "libmp3lame"}, false, false, false},
-		"mp3_rg":   {"mp3", 128, []string{"-c:a", "libmp3lame"}, true, false, false},
-		"opus":     {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, false, false, false},
-		"opus_rg":  {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, true, false, false},
-		"opus_car": {"opus", 96, []string{"-c:a", "libopus", "-vbr", "on"}, true, true, true},
+		"mp3": {
+			Format:  "mp3",
+			Bitrate: 128,
+			Options: []string{"-c:a", "libmp3lame"},
+		},
+		"mp3_rg": {
+			Format:     "mp3",
+			Bitrate:    128,
+			Options:    []string{"-c:a", "libmp3lame"},
+			ReplayGain: RGForce,
+		},
+		"opus": {
+			Format:  "opus",
+			Bitrate: 96,
+			Options: []string{"-c:a", "libopus", "-vbr", "on"},
+		},
+		"opus_rg": {
+			Format:     "opus",
+			Bitrate:    96,
+			Options:    []string{"-c:a", "libopus", "-vbr", "on"},
+			ReplayGain: RGForce,
+		},
+		"opus_car": {
+			Format:     "opus",
+			Bitrate:    96,
+			Options:    []string{"-c:a", "libopus", "-vbr", "on"},
+			ReplayGain: RGHigh,
+			Upsample:   true,
+		},
 	}
 }
 
@@ -96,35 +127,34 @@ func ffmpegCommand(filePath string, profile Profile) (*exec.Cmd, error) {
 		"-vn",
 		"-b:a", fmt.Sprintf("%dk", profile.Bitrate),
 	}
-	args = append(args, profile.ffmpegOptions...)
-	if profile.forceRG {
-		aBaselineGain := 6
-		if profile.hiGainRG {
-			// This baseline gain results in final track being +3~5dB louder
-			// than Foobar2000's default ReplayGain target volume.
-			// This makes it easier to listen to music in a car, where all other
-			// sources are usually ten thousand times louder than RG-adjusted music.
-			// -- @spijet
-			aBaselineGain = 15
-		}
-		aFilters := []string{
-			fmt.Sprintf("volume=replaygain=track:replaygain_preamp=%ddB:replaygain_noclip=0", aBaselineGain),
-			"alimiter=level=disabled",
-			"asidedata=mode=delete:type=REPLAYGAIN",
-		}
+	args = append(args, profile.Options...)
 
-		// opus always forces output to 48kHz sampling rate, but we can still use upsampling
-		// to increase RG and alimiter's peak limiting precision, which is desirable in some
-		// cases. ffmpeg's `soxr` resampler is quite fast on x86-64: it takes around 5 seconds
-		// on my Ryzen 3600 to transcode an 8-minute FLAC with 2x upsample and RG applied.
+	var aFilters []string
+	// opus always forces output to 48kHz sampling rate, but we can still use upsampling
+	// to increase RG and alimiter's peak limiting precision, which is desirable in some
+	// cases. ffmpeg's `soxr` resampler is quite fast on x86-64: it takes around 5 seconds
+	// on my Ryzen 3600 to transcode an 8-minute FLAC with 2x upsample and RG applied.
+	// -- @spijet
+	if profile.Upsample {
+		aFilters = append(aFilters, "aresample=96000:resampler=soxr")
+	}
+
+	switch profile.ReplayGain {
+	case RGForce:
+		aFilters = append(aFilters, ffmpegPreamp(6)...)
+	case RGHigh:
+		// This baseline gain results in final track being +3~5dB louder
+		// than Foobar2000's default ReplayGain target volume.
+		// This makes it easier to listen to music in a car, where all other
+		// sources are usually ten thousand times louder than RG-adjusted music.
 		// -- @spijet
-		if profile.upsample {
-			aFilters = append([]string{"aresample=96000:resampler=soxr"}, aFilters...)
-		}
-		aFilterString := strings.Join(aFilters, ", ")
+		aFilters = append(aFilters, ffmpegPreamp(15)...)
+	}
+
+	if len(aFilters) > 0 {
 		args = append(args,
 			// set up replaygain processing
-			"-af", aFilterString,
+			"-af", strings.Join(aFilters, ", "),
 			// drop redundant replaygain tags
 			"-metadata", "replaygain_album_gain=",
 			"-metadata", "replaygain_album_peak=",
@@ -134,6 +164,7 @@ func ffmpegCommand(filePath string, profile Profile) (*exec.Cmd, error) {
 			"-metadata", "r128_track_gain=",
 		)
 	}
+
 	args = append(args, "-f", profile.Format, "-")
 	ffmpegPath, err := exec.LookPath(ffmpeg)
 	if err != nil {
@@ -144,6 +175,14 @@ func ffmpegCommand(filePath string, profile Profile) (*exec.Cmd, error) {
 	// but please do let me know if you see otherwise
 }
 
+func ffmpegPreamp(dB int) []string {
+	return []string{
+		fmt.Sprintf("volume=replaygain=track:replaygain_preamp=%ddB:replaygain_noclip=0", dB),
+		"alimiter=level=disabled",
+		"asidedata=mode=delete:type=REPLAYGAIN",
+	}
+}
+
 func encode(out io.Writer, trackPath, cachePath string, profile Profile) error {
 	// prepare cache part file path
 	cachePartPath := fmt.Sprintf("%s.part", cachePath)
```
image

spijet: Yep, looking good!
As for `len(aFilters) > 0` check — it seems sufficient for now, because we only use audio filters when we're forcing ReplayGain, but it may lead to `ffmpeg` stripping out RG tags if/when we add some encoding profile with other filters in it (say, a profile that does on-the-fly stereo⇒binaural conversion for headphone listening).

What if we set it up like this?
```golang
var stripRGTags bool
switch profile.ReplayGain {
case RGForce:
    stripRGTags = true
...
}

if len(aFilters) > 0 {
    args = append(args, "-af", strings.Join(aFilters, ", "))
    if stripRGTags {
        args = append(args, "-metadata", ...)
    }
}
```

sentriz: ah good shout, thanks!
then in that case, as we want to add more functionality to the ffmpegCommand function, it might get a bit complicated in future.
what about a "functional options" pattern - where we can centralise the the different behaviours of each profile - and possibly make adding future functionality easier?

it could look something like this
image


https://termbin.com/p7ee

(not at all tested, but just wondering if you like this idea)

spijet: I was thinking about it too, but then I thought about a reverse approach, similar to what they do in Subsonic/MadSonic/AirSonic, where all encoding profiles are just four-field maps (converted to YAML for readability):
```yaml
# Simple example, FLAC → MP3
- sourceFormat: flac
  outputFormat: mp3
  # Format tokens: %s is source file, %f is output format, %b is bitrate
  stageOne: "ffmpeg -i %s -f %f -b:a %bk pipe:1"
  stageTwo:
# Two-stage example, some tracker module format → MP3
- sourceFormat: s3m
  outputFormat: mp3
  stageOne: "xmp --nocmd --stdout --bits 16 --frequency 44100 %s"
  stageTwo: "ffmpeg -i pipe:0 -f %f -b:a %bk pipe:1"
```

These allow for much more flexibility, but IMO are harder for end-users to make sense of. What do you think?

sentriz: I would be in favour of this :+1: seems nice a simple

though if it were a real yaml file, I'd be hesitant to add a second configuration file to gonic. 
maybe an idea could be to store the profiles as structs first? then later to make them configurable my the web UI. this would help with #121

spijet: You mean storing them in gonic's database? That sounds good, yeah. As for configuration files — a YAML file could be used as a "preseed", so the users could feed their profiles into gonic during the first deploy when using Docker.

sentriz:
howdy!
I just added these commits to master, I thought it would close the PR but it didn't 
image

I also added that enum stuff in on top, hope this is OK for you
here are some resulting args after testing

```
mp3      : -v 0 -i "03.12 Regiment.flac" -map 0:a:0 -vn -b:a 96k -c:a libmp3lame -f mp3 -
mp3_rg   : -v 0 -i "03.12 Regiment.flac" -map 0:a:0 -vn -b:a 96k -c:a libmp3lame -af volume=replaygain=track:replaygain_preamp=6dB:replaygain_noclip=0, alimiter=level=disabled, asidedata=mode=delete:type=REPLAYGAIN -metadata replaygain_album_gain= -metadata replaygain_album_peak= -metadata replaygain_track_gain= -metadata replaygain_track_peak= -metadata r128_album_gain= -metadata r128_track_gain= -f mp3 -
opus     : -v 0 -i "03.12 Regiment.flac" -map 0:a:0 -vn -b:a 96k -c:a libopus -vbr on -f opus -
opus_rg  : -v 0 -i "03.12 Regiment.flac" -map 0:a:0 -vn -b:a 96k -c:a libopus -vbr on -af volume=replaygain=track:replaygain_preamp=6dB:replaygain_noclip=0, alimiter=level=disabled, asidedata=mode=delete:type=REPLAYGAIN -metadata replaygain_album_gain= -metadata replaygain_album_peak= -metadata replaygain_track_gain= -metadata replaygain_track_peak= -metadata r128_album_gain= -metadata r128_track_gain= -f opus -
opus_car : -v 0 -i "03.12 Regiment.flac" -map 0:a:0 -vn -b:a 96k -c:a libopus -vbr on -af aresample=96000:resampler=soxr, volume=replaygain=track:replaygain_preamp=15dB:replaygain_noclip=0, alimiter=level=disabled, asidedata=mode=delete:type=REPLAYGAIN -f opus -
```

if it looks good to you, I might make a release shortly - then we could do cool stuff with user configurable profiles from the admin UI at a later stage. thoughts? thanks!

PR #159, opened 2021-10-04
feat: add CreatedAt to albums

Duncaen:
This change will make the GetAlbumList views with the "newest" type
work as expected.


sentriz:
thanks! looks good

Issue #177, opened 2021-12-15
Scan for 'artist.jpg' or 'cover.jpg' to update artist thumbnail


ZachMyers3:
Currently I believe the default last.fm thumbnail is filled. Similar to album art it would be nice to scrape for a "cover.jpg" or "artist.jpg" in the artist directory and update off that if it's there.

[Here](http://www.subsonic.org/pages/api.jsp#getArtistInfo) is the reference to Subsonic API endpoint for reference.

ZachMyers3:
Taken care of with #178

PR #178, opened 2021-12-18
Arbitrarily grab artist cover art

ZachMyers3:
For #177 

My work so far to grab that artist artwork while scanning and save it to the artist information. Currently artist information is inserted through track metadata, so this implementation assumes that the parent folder of the current traversed album folder is the artist directory. It also assumes that the artist directory will have the artwork if it's there. If both of these assumptions are met then it populates the Artist database with the absolute path to the image.

Any input would be appreciated, this is my first time working with Go!

Thanks

### DIFF
--- a/server/ctrlsubsonic/handlers_by_folder.go
+++ b/server/ctrlsubsonic/handlers_by_folder.go
@@ -245,3 +245,13 @@ func (c *Controller) ServeSearchTwo(r *http.Request) *spec.Response {
 func (c *Controller) ServeGetArtistInfo(r *http.Request) *spec.Response {
 	return spec.NewResponse()
 }
+
+func (c *Controller) ServeGetStarred(r *http.Request) *spec.Response {
+	sub := spec.NewResponse()
+	sub.Starred = &spec.Starred{
+		Artists: []*spec.Directory{},
+		Albums:  []*spec.TrackChild{},
+		Tracks:  []*spec.TrackChild{},
+	}
+	return sub
+}
--- a/server/ctrlsubsonic/handlers_by_tags.go
+++ b/server/ctrlsubsonic/handlers_by_tags.go
@@ -361,3 +361,13 @@ func (c *Controller) ServeGetSongsByGenre(r *http.Request) *spec.Response {
 	}
 	return sub
 }
+
+func (c *Controller) ServeGetStarredTwo(r *http.Request) *spec.Response {
+	sub := spec.NewResponse()
+	sub.StarredTwo = &spec.StarredTwo{
+		Artists: []*spec.Artist{},
+		Albums:  []*spec.Album{},
+		Tracks:  []*spec.TrackChild{},
+	}
+	return sub
+}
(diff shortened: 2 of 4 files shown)


Duncaen:
This should probably use the `getCoverArt` endpoint instead of inventing one as this is how its defined in the subsonic api:

> The ID of a song, album or artist.

http://www.subsonic.org/pages/api.jsp#getCoverArt

ZachMyers3:
Good catch! I wasn't aware. I'll make the change.

The [ArtistInfo2](http://www.subsonic.org/pages/api.jsp#getArtistInfo2) endpoint references the URL of the images for an artist, do you have any suggestion on how to serve the URL from within the server? `baseURL` wouldn't work behind a reverse proxy.

Thanks for the help

Duncaen:
> The [ArtistInfo2](http://www.subsonic.org/pages/api.jsp#getArtistInfo2) endpoint references the URL of the images for an artist, do you have any suggestion on how to serve the URL from within the server? `baseURL` wouldn't work behind a reverse proxy.

Hmm yea no idea other than adding a flag to set a base url, clients seem to expect a full url.

sentriz:
hi! thanks for your work on this and sorry it's taken a while to have a look :) 

i like the idea. also unrelated - I wonder if beets.io has a plugin to populate these artworks automatically? or how have you done is so far?

as for gonic, you may know every folder on the filesystem has a respective album record which are only sometimes shown or rendered differently based on if we browse by tags or not

so I think there may be a simpler approach without any scanner, walking, database etc changes 


in ServeGetArtistInfoTwo there could be 

```golang
var artistFolder db.Album
c.DB.
	Select("parent.*").
	Joins("JOIN albums parent ON parent.id=albums.parent_id").
	Where("albums.tag_artist_id=?", id.Value).
	Find(&artistFolder)
```

then  there could be logic like `if artistFolder.Cover != "", then cover URL is "https://gonic/getCoverArt?id=al-123"`

the only different I see with that compared to these changes it that the cover file in the artist's folder would need to be cover.jpg, folder.jpg, etc instead of artist.jpg (but we could also add artist.* as an acceptable name)

interested to hear your thoughts on this. thanks!

ZachMyers3:
To populate [I wrote a little python script](https://github.com/ZachMyers3/fanarttv_artist_art) that's pretty specific to my library to query FanartTV for artist thumbnails. 

I appreciate the input, I wasn't aware of how the `parent` schema worked and it seems like the better option. I'll work on implementing this and see how it goes. The input is much appreciated!

ZachMyers3:
That seemed to work, I just need to figure out the best way to pass a server url configuration to that endpoint and it should be good.

ZachMyers3:
I believe I have this working the way @sentriz suggested, and I'm passing a server url through the controller to get a proper url within the `GetArtistInfo2` route.

Let me know what you think!

ZachMyers3:
With the help from @sentriz I think we're good to go now.

Thanks for helping me out! 😃

sentriz:
hey I rebased this branch and for some reason github closesed the PR.. not sure. anyway I just pushed this feature to master. 

thanks for this feature! very cool :)

Issue #179, opened 2021-12-27
Add coverArt to getArtists response


ZachMyers3:
[Example here](http://www.subsonic.org/pages/inc/api/examples/artists_example_1.xml)

Some clients depends on the `coverArt` response from the client to populate an artists' cover art. Now that gonic is capable of displaying cover art for an artist it should be possible to display it at this endpoint.

PR #180, opened 2021-12-27
adjust artist by tags to hunt for artist cover

ZachMyers3:
For #179 

Propagate the logic we've used for artist covers into other endpoints. This should cover `getArtsits` `getArtist` and `ServeSearchThree` now as well.

Only displays a cover art tag if we have it locally.

I believe when the rebase took place `coverGetPathArtist` didn't make it into the merge as well, so I've corrected that.

### DIFF
--- a/server/ctrlsubsonic/handlers_by_tags.go
+++ b/server/ctrlsubsonic/handlers_by_tags.go
@@ -44,7 +44,7 @@ func (c *Controller) ServeGetArtists(r *http.Request) *spec.Response {
 			resp = append(resp, indexMap[key])
 		}
 		indexMap[key].Artists = append(indexMap[key].Artists,
-			spec.NewArtistByTags(artist))
+			spec.NewArtistByTags(artist, c.DB))
 	}
 	sub := spec.NewResponse()
 	sub.Artists = &spec.Artists{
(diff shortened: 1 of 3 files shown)


ZachMyers3:
Is there anything needed from me? As of right now the `nightly` branch is non-working for artist-based queries by tags because of the lack of `coverGetPathArtist` existing in the current branch.

sentriz:
howdy! thanks for this, I also forgot about the artist view's cover prop. 

my thinking about those "spec construct" functions is that they should be thin conversions from the database representation to the subsonic representation of some data.  so doing some database lookups in those functions would be a new idea

on top of that, if we need to do this lookup in many places, we should probably store it (similar to your previous pr! :eyes:)

I added a commit https://github.com/sentriz/gonic/commit/c0ebd2642206f7dba81f136cd9d576ded75bb14e that stores the guessed artist folder as a column in the artists table. that way the folder can be shared in not only the getArtistInfo lookup as we had already, but also the search3 and getArtists as you suggested :+1:

I did some testing myself and it seems to work, please let me know if not  

PS sorry I couldn't to a PR review I very little time this week

sentriz:
PS! for the meantime, the change I added requires at least one track to be `touch`ed if it's respective artists cover file has changed 

eg

```
/music/The Fall/artist.png                              # if this was added on its own after the music, then
/music/The Fall/Dragnet (1979)/Psykick Dancehall.flac   # touch this afterwards for above to be picked up
```

and that's because the artist table is traditionally populated only from audio tags. it would be a bit complicated to change this I think. but let me know if you think it's worth changing

ZachMyers3:
This all works great! I think the way it's implemented now (seeking a track to populate an artist) works fine the way it is and that was how I was populating covers with my initial implementation as well.

The only thing I noticed is the `nightly` tag for the docker image doesn't have this commit, but pulling and running locally it worked as expected.

Thanks for the help!

Issue #135, opened 2021-05-05
Sort getAlbumList "newest" / getAlbumList2 "newest"  / admin UI folders by ctime 


bobberb:
Dear sentriz,

Thank you for an excellent lightweight alternative. 

I have recently deployed gonic alongside airsonic and I've noticed few small cosmetic issues:

1. Some album art, if not labeled exactly as cover.jpg or front.jpg or AlbumArt.jpg or equivalent will be missing. How do I trigger a refresh of just the album art?
2. After renaming "missing" album art, the albums move to the top of the recent folders list. I sort my music mentally by when the folder was created. Gonic updates the "recent albums" ?based on the last modtime of any file within? (such as a renamed album art image). Is there a way to avoid this?

jef:
> Some album art, if not labeled exactly as cover.jpg or front.jpg or AlbumArt.jpg or equivalent will be missing. How do I trigger a refresh of just the album art?

The Subsonic API doesn't support this, so I don't know if this is possible (as of now). You'd have to do a full scan.

> After renaming "missing" album art, the albums move to the top of the recent folders list. I sort my music mentally by when the folder was created. Gonic updates the "recent albums" ?based on the last modtime of any file within? (such as a renamed album art image). Is there a way to avoid this?

It is sorted by modified date:

https://github.com/sentriz/gonic/blob/e9cc2e8ea610686396e27645f3d12dde4ac92baf/server/assets/pages/home.tmpl#L119-L124

https://github.com/sentriz/gonic/blob/e9cc2e8ea610686396e27645f3d12dde4ac92baf/server/ctrladmin/handlers.go#L68-L72

https://github.com/sentriz/gonic/blob/6a99cdf299e0b9f2c3254786e19f9e56e8006121/server/scanner/scanner.go#L379-L382

If you wanted to use "created at" or something similar, this would have to be changed to be dynamic. Not sure if that's something that gonic has in scope though, as it's not really a viewing tool, but rather a server to host the Subsonic API to have tools like https://airsonic.netlify.app/ serve the media.

sentriz:
hey, yeah that's a strange one. that code was written a good while ago but i don't remember why it was implemented that way

looking at some docs now, and it does seem that ctime time is not exposed as simply as mtime is

also for the album art thing - one trick would be to `$ touch album_dir/` then do a quick incremental scan

bobberb:
[...]

PR #181, opened 2021-12-29
Expanded file date options and added Cover art scanning

b-levin:
For [Issue #135](https://github.com/sentriz/gonic/issues/135)

Expanded the date parameters stored in the `Album` struct to include `CreatedAt` (in addition to `ModifiedAt`). Additionally, users can now use the `create-sort` option in the file config or command line flag to change which date (`CreatedAt` or `ModifiedAt`) is used for the sorting in the `ServeHome` function.

The ability to just scan for album cover art changes was added to mirror the feature from airsonic. Users can manually trigger the cover art scan from the homepage by clicking on the new `cover scan` button under recent folders on the admin/home page. Additionally, a ticker was added for the new cover art scan, which is disabled by default and can be modified via either the `cover-scan-interval` in the config file or command line flag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ dist
 gonic
 gonicscan
 gonicembed
+.vscode
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
-FROM golang:1.16-alpine AS builder
+FROM golang:1.17-alpine AS builder
 RUN apk add -U --no-cache \
-  build-base \
-  ca-certificates \
-  git \
-  sqlite \
-  taglib-dev \
-  alsa-lib-dev \
-  zlib-dev \
-  go
+    build-base \
+    ca-certificates \
+    git \
+    sqlite \
+    taglib-dev \
+    alsa-lib-dev \
+    zlib-dev \
+    go
 WORKDIR /src
 COPY go.mod .
 COPY go.sum .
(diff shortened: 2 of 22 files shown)


sentriz:
Hi and thanks. I'm not sure what the benefit of scanning only covers is. Is it a performance thing? It seems to me like a lot of added complexity

b-levin:
The idea of the cover scan was indeed performance related and to address the first point @bobberb made in their issue (#135). While it does add complexity to the code base, it is significantly faster than doing a full scan (seconds as opposed to minutes). If the added complexity is too much, I can redo the PR to just include the additional date options.

sentriz:
Thanks for the context on cover scanning. Personally on my system a quick scan takes about 15s on 50k tracks and will pick up any new or changed covers. May I ask with your system, how many tracks? And where is the db file stored? Spinny disk or SSD? Also, is the music mounted locally, or a remote mount? Cheers

bobberb:
@sentriz 

I have pulled and built this PR multiple times over the last 24 hours. 

Since I opened issue #135, unfortunately, airsonic already failed and was forked by another team. It still remains a very well-running docker image, albeit one that soaks up 2GB RAM at idle while your nice goland gonic does not. 

I have been unable to fully commit to Gonic because moving and touching those files in the folder of an album could have reset it's position in Recently Added.

I speak as a migrant from Airsonic, I much prefer having a stable "Recently Added" list, especially one that can tolerate album art changes. Current PR allows me to drop in album art without changing album order in that list. I've been running both your code and Airsonic side by side for a good while now. I much prefer gonic for speed and memory cost however my mind is still oriented to "Recently Added = Recently Downloaded order".

I agree with @b-levin that it should be based on a number given by the filesystem - not gonic. 

@b-levin thank you!

bobberb:
I should note that at the current time, this branch does not currently respect albums which are present prior to gonic. 1st import will be random order seemingly. Every album dropped in thereafter is accepted in a list which is immune from movement by album art updates.

bobberb:
> Thanks for the context on cover scanning. Personally on my system a quick scan takes about 15s on 50k tracks and will pick up any new or changed covers. May I ask with your system, how many tracks? And where is the db file stored? Spinny disk or SSD? Also, is the music mounted locally, or a remote mount? Cheers

I got 7s for both scans - I'm myself more excited for a stable list on this branch.

b-levin:
I was predominately leaning on @bobberb for performance testing, as my library is too small to have a real difference in testing. The more stable list is the highlight of these changes. I'm only working with a few hundred tracks on locally mounted SSD storage, plus when combined with the fact I've been testing these changes on my desktop results in both cover and quick scans to take only a second or two at the most. 

Given the fact that the cover scan seems to be a bit redundant, would the best approach be to remove it from this PR and keep the additional date?

sentriz:
howdy and sorry for the delay

yeah @b-levin I think we could remove the cover scanning feature for now - then it will be easy for me to review the changes around ctime/mtime


after that, we could have a look at cover scanning again as you suggest, or else see if the scanner performance could be improved in the case of only covers changing. a bit more easy for the user 

thanks!

b-levin:
@sentriz Not a problem, this can definitely be a busy time of year!

Given that the cover scan feature is getting split out, I'm going to close this PR and open and open a new one just with the time changes.

PR #186, opened 2022-01-21
Updated README.md to clarify directory structure

### LINKED ISSUE #185: Folder layout requirements unclear
I'm a little confused on what the required directory structure is.  According to the README, an album must be in a single folder, but according to #89, a folder must contain files from only one album.  What is the actual requirement?  Is it both?  Whatever the requirement, I'm happy to submit a pull request that clarifies the text in the README.

jgdye:
In reference to #185, clarifies that all files from an album must be in the same directory and the directory must only contain files from that album.

fixes #185

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,9 @@ queries like show me "recently played compilations" or "recently added albums" a
 
 ## directory structure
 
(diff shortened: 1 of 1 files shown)


sentriz:
nice thanks!

PR #192, opened 2022-02-06
Show if a scan is running.

djl:
Currently there's no way to tell if a scan is running without looking at the logs. Now there is 😄


sentriz:
cool thanks!

PR #195, opened 2022-02-09
Feat/get similar songs 2, Feat/get similar songs, Feat/get top songs

xaviercrochet:
Implementation of subsonic [getSimilarSongs2](www.subsonic.org/pages/api.jsp#getSimilarSongs2) and Last.fm [artist.getSimilar](https://www.last.fm/api/show/artist.getSimilar)

N.B. I inspired myself of [ampache](https://github.com/ampache/ampache/) 's implementation `src/Module/Api/Subsonic_Api.php:2229` as Last.Fm doesn't expose such endpoint.

The logic is the following:

1. Retrieve artist 's similar artists using Last.fm 's artist.getSimilar
2. Look in our database for tracks matching the similar artists
3. Randomize (then trim) the track set to prevent ending up with X track of the same artist. 

What do you think about this approach?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ gonic
 gonicscan
 gonicembed
 .vscode
+*.swp
--- a/server/ctrlsubsonic/handlers_by_tags.go
+++ b/server/ctrlsubsonic/handlers_by_tags.go
@@ -407,15 +407,14 @@ func (c *Controller) genAlbumCoverURL(r *http.Request, folder *db.Album, size in
 }
 
 func (c *Controller) ServeGetTopSongs(r *http.Request) *spec.Response {
-
 	params := r.Context().Value(CtxParams).(params.Params)
-	sub := spec.NewResponse()
 	count := params.GetOrInt("count", 10)
 	artist, err := params.Get("artist")
 	if err != nil {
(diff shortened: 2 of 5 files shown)


sentriz:
just had a look and I had a few minor comments. mostly around formatting and I think some stuff wasn't needed

how do you feel about these changes?
https://github.com/sentriz/gonic/commit/99c2022eac0e2149ecf3cb7a5c7947e9a874105d

to summarise
- shuffing the results at the database level instead of application, using the clause LIMIT clause based on the size param. I think this should not have adjacent tracks by the same artist like you mention
- updating the lastfm API helper functions to not depend on the db package
- moving the spec struct variables to the bottom of the handlers

how does it sound to you? does it work still? I tested a bit and it seems to 

thank you kindly!

xaviercrochet:
Sounds good, thanks!

xaviercrochet:
Done!

sentriz:
hey I split the changes into 3 commits, one for each new view, and added you as the author. this way it shows up nicely in the changelog for the next release 

https://github.com/sentriz/gonic/pull/160

let me know if all is working well still. thanks very much!

xaviercrochet:
Thank you @sentriz.

Have a nice one!

PR #196, opened 2022-02-09
fast artist cover scanning

sentriz:
- revert guessed artist folder changes
- feat: render local artist images with no foreign key
- chore: upgrade deps


sentriz:
hey @ZachMyers3 I think the current implementation of the artist covers is causing scans to be slower, maybe because of some cyclical foreign keys. this PR reverts that and does it again in a way that is more similar to your original implementation :+1: 

I tested this myself and seems to work well, let me know if you find anything though. thanks!

ZachMyers3:
Works for me as well. Thanks for letting me know. Surprised it slows the scans down so much.

PR #201, opened 2022-02-24
Support public playlists

grilix:
When multiple people share the same instance, they might want to share
their playlists between them.

This allows people to mark playlists as public, and to listen to public
playlists from other people. Listeners will also know who owns the
playlist, to help avoid confusion and make this feature a bit nicer.

Subsonic restrict updating playlists only to owners, this honors that
behavior, but adding flexibility could be achieved easily.

### DIFF
--- a/server/ctrlsubsonic/handlers_playlist.go
+++ b/server/ctrlsubsonic/handlers_playlist.go
@@ -23,6 +23,8 @@ func playlistRender(c *Controller, playlist *db.Playlist) *spec.Playlist {
 		Comment:   playlist.Comment,
 		Created:   playlist.CreatedAt,
 		SongCount: playlist.TrackCount,
+		Public:    playlist.IsPublic,
+		Owner:     user.Name,
 	}
 
 	trackIDs := playlist.GetItems()
@@ -48,7 +50,7 @@ func playlistRender(c *Controller, playlist *db.Playlist) *spec.Playlist {
 func (c *Controller) ServeGetPlaylists(r *http.Request) *spec.Response {
 	user := r.Context().Value(CtxUser).(*db.User)
 	var playlists []*db.Playlist
-	c.DB.Where("user_id=?", user.ID).Find(&playlists)
+	c.DB.Where("user_id=?", user.ID).Or("is_public=?", true).Find(&playlists)
 	sub := spec.NewResponse()
 	sub.Playlists = &spec.Playlists{
 		List: make([]*spec.Playlist, len(playlists)),
@@ -90,6 +92,9 @@ func (c *Controller) ServeCreatePlaylist(r *http.Request) *spec.Response {
 		FirstOrCreate(&playlist)
 
 		// update meta info
+	if playlist.UserID != 0 && playlist.UserID != user.ID {
+		return spec.NewResponse()
+	}
 	playlist.UserID = user.ID
 	if val, err := params.Get("name"); err == nil {
 		playlist.Name = val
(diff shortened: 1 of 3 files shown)


grilix:
Made some last minute changes. It's now working, and the docker image is available at [grilix/gonic:public-playlists](https://hub.docker.com/layers/grilix/gonic/public-playlists/images/sha256-308bb6c108ad8ada99ba07c7dbf163126fa19001aa2aea225459c26d5c00ac25) if needed.

Let me know what you think.

server/ctrlsubsonic/handlers_playlist.go
```diff
@@ -123,13 +133,22 @@ func (c *Controller) ServeUpdatePlaylist(r *http.Request) *spec.Response {
...
 
 		// update meta info
-	playlist.UserID = user.ID
+	if playlist.UserID == 0 {
+		playlist.UserID = user.ID
+	} else {
+		if playlist.UserID != user.ID {
+			return spec.NewResponse()
+		}
+	}
```
sentriz, line 142: could we just do this?

```suggestion
	if playlist.UserID != 0 && playlist.UserID != user.ID {
		return spec.NewResponse()
	}
	playlist.UserID = user.ID
```

server/ctrlsubsonic/handlers_playlist.go
```diff
@@ -90,10 +93,17 @@ func (c *Controller) ServeCreatePlaylist(r *http.Request) *spec.Response {
...
```
sentriz, line 102: could we just do this?

```suggestion
	if playlist.UserID != 0 && playlist.UserID != user.ID {
		return spec.NewResponse()
	}
	playlist.UserID = user.ID
```

server/ctrlsubsonic/handlers_playlist.go
```diff
@@ -90,10 +93,17 @@ func (c *Controller) ServeCreatePlaylist(r *http.Request) *spec.Response {
...
 	if val, err := params.Get("name"); err == nil {
 		playlist.Name = val
 	}
+	playlist.IsPublic = false
```
sentriz, line 106: in go, the zero value of bools is "false" (and you can't have a nil bool)

so we could say, if we like:

```suggestion
```

server/ctrlsubsonic/handlers_playlist.go
```diff
@@ -123,13 +133,22 @@ func (c *Controller) ServeUpdatePlaylist(r *http.Request) *spec.Response {
...
+	}
 	if val, err := params.Get("name"); err == nil {
 		playlist.Name = val
 	}
 	if val, err := params.Get("comment"); err == nil {
 		playlist.Comment = val
 	}
+	if val, err := params.Get("public"); err == nil {
+		playlist.IsPublic = strings.ToLower(val) == "true"
+	}
```
sentriz, line 151:
```suggestion
	if val, err := params.GetBool("public"); err == nil {
		playlist.IsPublic = val
	}
```

sentriz (commented):
thanks! this is looking good!

since we're changing the db schema, we need to write a migration so that everyone's database gets updated

they are in here
server/db/migrations.go

and since it's a simple change (adding one column) gorm can do it for us with `tx.AutoMigrate(Playlist{}).Error`

grilix:
@sentriz great suggestions, it looks way cleaner now!

I initially had a migration but was broken, so I though `gorm` was automatically handling that and removed it. I've now added one that seems to work.

sentriz:
tested locally and seems to work well. thank you kindly :)

PR #203, opened 2022-02-25
Improve how subsonic ServceGatArtistInfoTwo is retrieving similar artist

xaviercrochet:
Hey there

I made the following improvement to ```/getArtistInfo2```

I realized that, as  ```lastfm.artist.getInfo``` as only retrieving a couple of artist, ```/getArtistInfo2``` was most of the time returning very few (or no) similar artists. 

- Retrieve similar artists using ```lastfm.artist.getSimilar``
- Match similar artists result with internal database in one single sql query, using a similar pattern as we used for ``` ServeGetSimilarSongs``` and ```ServeGetSimilarSongsTwo```

### DIFF
--- a/server/ctrlsubsonic/handlers_by_tags.go
+++ b/server/ctrlsubsonic/handlers_by_tags.go
@@ -296,31 +296,41 @@ func (c *Controller) ServeGetArtistInfoTwo(r *http.Request) *spec.Response {
 
 	count := params.GetOrInt("count", 20)
 	inclNotPresent := params.GetOrBool("includeNotPresent", false)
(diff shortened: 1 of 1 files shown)


sentriz:
cool thanks!

sentriz:
actually I just realised that with this change we never return any artists that aren't in the database (not respecting the includeNotPresent param) so I adjusted the commit to be a bit more similar to how it was, just using the lastfm.ArtistGetSimilar

does that seem ok to you?

image

xaviercrochet:
Yes this will do the trick, thanks!

PR #206, opened 2022-03-14
Added https support. Will revert to http if either cert or key are em…

brian-doherty:
…pty strings.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -276,9 +276,8 @@ func (s *Server) StartHTTP(listenAddr string, tlsCert string, tlsKey string) (Fu
 			log.Print("starting job 'http'\n")
 			if tlsCert != "" && tlsKey != "" {
 				return list.ListenAndServeTLS(tlsCert, tlsKey)
-			} else {
-				return list.ListenAndServe()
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -274,7 +274,11 @@ func (s *Server) StartHTTP(listenAddr string) (FuncExecute, FuncInterrupt) {
 	}
 	return func() error {
 			log.Print("starting job 'http'\n")
-			return list.ListenAndServe()
+			if ((len(tlsCert) > 0) && (len(tlsKey) > 0)) {
```
sentriz, line 277:
```suggestion
			if tlsCert != "" && tlsKey != "" {
```

sentriz (commented):
thanks! looks good :+1: just one thing

server/server.go
sentriz: it's a bit more common to compare against string literals in go

sentriz:
also seems we have 3 small lint issues
image

sentriz:
thanks, just updated the docs too :+1: 

image

PR #207, opened 2022-03-17
Added update of play time and count when scrobbling as per API docs

brian-doherty:


sentriz:
thanks! I also wanted to add some error checking to those functions so I included that too

PR #213, opened 2022-04-19
Added error case for when DownloadEpisode is called via API and podcast is already downloaded

brian-doherty:


sentriz:
@lxea thoughts on this one

lxea:
This seems fine to me, sometimes it can be useful to redownload stuff when a response is an error page which was happening to me with some feed hosts, this should probably just be handled properly by trying to parse the mp3 or checking the staus code or something

brian-doherty:
FYI without this fix I was getting multiple copies of the same podcast with different filenames:

image

sentriz:
thanks :+1: squashed and added

@lxea does this mean that we don't need the rename `(1)` `(2)` logic now?

PR #214, opened 2022-04-21
Added getNewestPodcasts API

brian-doherty:


db/model.go
```diff
@@ -371,6 +371,10 @@ func (pe *PodcastEpisode) SID() *specid.ID {
 	return &specid.ID{Type: specid.PodcastEpisode, Value: pe.ID}
 }
 
+func (pe *PodcastEpisode) PID() *specid.ID {
+	return &specid.ID{Type: specid.PodcastEpisode, Value: pe.PodcastID}
```
sentriz, line 375:
```suggestion
func (pe *PodcastEpisode) PodcastSID() *specid.ID {
	return &specid.ID{Type: specid.Podcast, Value: pe.PodcastID}
```

sentriz (commented):
looks pretty good to me, but I think the specid should be Podcast?

brian-doherty:
You are correct. My mistake!

sentriz:
looks like we just need to update the calls to PID()


image

brian-doherty:
OK sorted.

sentriz:
lookin good to me, cheers brian :+1: 

image

sentriz:
btw i just noticed this

image

has it always been like this? :grin: never noticed

brian-doherty:
LOL yeah. UX is clearly not my jam because I just let stuff like that slide by me.

PR #221, opened 2022-05-06
Fixed rendering of podcast section on admin/home for both desktop and mobile

brian-doherty:
Fixed rendering of podcast section on admin/home for both desktop and mobile


sentriz:
cheers

PR #228, opened 2022-06-25
Avatar support

brian-doherty:


brian-doherty:
See what you think. I made delete red as a warning but of course that's easy to change.

brian-doherty:
Hey, so what do you think of my changes?

sentriz:
howdy! thanks very much for the changes . haven't had a chance to look yet but will shortly 👌

sentriz:
hiya brian! sorry it took so long to get back to you. not much free time recently and didn't have time for a full review  

but this looks really nice now, thanks very much :+1:  :+1: 

made some small changes, hope it's ok. let me know if not

- got rid of the inline <script />s this has casused issues with CSP in the past (unsafe-inline IIRC, not sure though) instead I just made the upload functionality the same as the playlist upload, using the external script file
- this also means the style of the upload button is the same (incl mouse over etc)
- some go styling stuff

PR #236, opened 2022-09-08
Add GONIC_HTTP_LOG setting to README

hufman:
There's an [undocumented http logging setting](https://github.com/sentriz/gonic/blob/7d0d036f0bc0f5bd4429f510ca636daa3934766b/cmd/gonic/gonic.go#L42) which I found useful to disable, and perhaps others might find it useful too.


sentriz:
thanks :+1:

PR #237, opened 2022-09-08
Add a ping endpoint that doesn't create a session

hufman:
I'm trying to reduce idle disk activity on my install, and so I want to configure my loadbalancer to check an endpoint that doesn't trigger a new session creation in the database.


sentriz:
nice one. though I think you can remove the .view suffix since its not a subsonic endpoint

hufman:
Can do!

sentriz:
thanks :+1:

PR #240, opened 2022-09-20
update musicbrainz metadata

tpeacock19:
Currently, the musicbrainz scrobbler is submitting incorrect data. The mappings for this api can be found [here](https://picard-docs.musicbrainz.org/downloads/MusicBrainz_Picard_Tag_Map.html):

```
Json Field      : MusicBrainz Internal Name	: Metadata Tag Name

release_mbid	: musicbrainz_releaseid		: musicbrainz_albumid
recording_mbid	: musicbrainz_recordingid	: musicbrainz_trackid
track_mbid	: musicbrainz_trackid		: musicbrainz_releasetrackid

```

Gonic submits the metadata tag `musicbrainz_trackid` (https://github.com/sentriz/gonic/blob/master/scanner/tags/tags.go#L41) as `track_mbid` (https://github.com/sentriz/gonic/blob/master/scrobble/listenbrainz/listenbrainz.go#L62). But that field actually corresponds to `musicbrainz_recordingid`.  The correct metadata field to use is `musicbrainz_releasetrackid`.

This PR corrects this and adds two more entries in the AdditionalInfo
struct, `RecordingMBID` and `ReleaseMBID`. Reflecting the mapping above.

### DIFF
--- a/db/model.go
+++ b/db/model.go
@@ -79,25 +79,27 @@ type AudioFile interface {
 }
 
 type Track struct {
-	ID             int `gorm:"primary_key"`
-	CreatedAt      time.Time
-	UpdatedAt      time.Time
-	Filename       string `gorm:"not null; unique_index:idx_folder_filename" sql:"default: null"`
-	FilenameUDec   string `sql:"default: null"`
-	Album          *Album
-	AlbumID        int `gorm:"not null; unique_index:idx_folder_filename" sql:"default: null; type:int REFERENCES albums(id) ON DELETE CASCADE"`
-	Artist         *Artist
-	ArtistID       int      `gorm:"not null" sql:"default: null; type:int REFERENCES artists(id) ON DELETE CASCADE"`
-	Genres         []*Genre `gorm:"many2many:track_genres"`
-	Size           int      `sql:"default: null"`
-	Length         int      `sql:"default: null"`
-	Bitrate        int      `sql:"default: null"`
-	TagTitle       string   `sql:"default: null"`
-	TagTitleUDec   string   `sql:"default: null"`
-	TagTrackArtist string   `sql:"default: null"`
-	TagTrackNumber int      `sql:"default: null"`
-	TagDiscNumber  int      `sql:"default: null"`
-	TagBrainzID    string   `sql:"default: null"`
+	ID                   int `gorm:"primary_key"`
+	CreatedAt            time.Time
+	UpdatedAt            time.Time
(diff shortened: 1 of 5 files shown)


sentriz:
hi! thanks for noticing this! i think this solution works well for new installations of gonic, but for existing installations that are upgraded, it would required migrating peoples databases. could get messy

how do you feel about a simpler solution, which is just to submit the recording_mbid to listenbrainz, coming from people's existing track BrainzIDs() ?

for example


```diff
diff --git a/scrobble/listenbrainz/listenbrainz.go b/scrobble/listenbrainz/listenbrainz.go
index 6f17167..73874d1 100644
--- a/scrobble/listenbrainz/listenbrainz.go
+++ b/scrobble/listenbrainz/listenbrainz.go
@@ -27,9 +27,10 @@ var (
 )
 
 type AdditionalInfo struct {
-	TrackNumber int    `json:"tracknumber,omitempty"`
-	TrackMBID   string `json:"track_mbid,omitempty"`
-	TrackLength int    `json:"track_length,omitempty"`
+	TrackNumber   int    `json:"tracknumber,omitempty"`
+	TrackMBID     string `json:"track_mbid,omitempty"`
+	RecordingMBID string `json:"recording_mbid,omitempty"`
+	TrackLength   int    `json:"track_length,omitempty"`
 }
 
 type TrackMetadata struct {
@@ -58,9 +59,9 @@ func (s *Scrobbler) Scrobble(user *db.User, track *db.Track, stamp time.Time, su
 	payload := &Payload{
 		TrackMetadata: &TrackMetadata{
 			AdditionalInfo: &AdditionalInfo{
-				TrackNumber: track.TagTrackNumber,
-				TrackMBID:   track.TagBrainzID,
-				TrackLength: track.Length,
+				TrackNumber:   track.TagTrackNumber,
+				RecordingMBID: track.TagBrainzID,
+				TrackLength:   track.Length,
 			},
 			ArtistName:  track.TagTrackArtist,
 			TrackName:   track.TagTitle,
```

reference:
https://picard-docs.musicbrainz.org/downloads/MusicBrainz_Picard_Tag_Map.html
https://listenbrainz.readthedocs.io/en/latest/users/json.html#json-doc

sentriz:
did some testing and it seems to work well, so added it to master for now. let me know if that's ok for you. thanks!

tpeacock19:
yeah, I did have to switch between the two database versions while I was doing this. I think submitting the recording would be fine. I would suggest including the album ID if possible or else there is no way to differentiate between versions of an album.

PR #243, opened 2022-10-01
Added support for star / rating.

brian-doherty:


sentriz:
thanks brian :+1: have you considered using gorm's Preload feature instead of the addStarRatingToXXX methods?

for example if you added here
https://github.com/sentriz/gonic/pull/243/files#diff-6e27daff183350ecbdf054691f655ee921cf775dcfdb4632e5981cd4d03338eaR90

`.Preload("TrackRating", "user_id=?", user.ID)`

it would just require that Track has a TrackRating pseudo column, just like Track.Album etc, not a real column

that's the current pattern anyway and we wouldn't need the addStartRatingToXXX calls

brian-doherty:
OK let me see what I can do.

brian-doherty:
Unsure how to handle the whole Directory / Album / TrackChild system that way since they could be one of multiple. Do you put in the union and check which is populated?

server/ctrlsubsonic/handlers_by_tags.go
```diff
@@ -380,18 +412,84 @@ func (c *Controller) ServeGetSongsByGenre(r *http.Request) *spec.Response {
...
+		if err := c.DB.Where("artist_id=?", a.TagArtistID).Find(&ar); err != nil {
+			return spec.NewError(0, "find artist for album: %v", err)
+		}
+		album := spec.NewAlbumByTags(a, &ar)
+		sub.StarredTwo.Albums = append(sub.StarredTwo.Albums, album)
+	}
+
+	// tracks
+	var tracks []*db.Track
+	q = c.DB.Table("tracks").Joins("right join track_stars on tracks.id=track_stars.track_id").
```
sentriz, line 474: the right joins are giving me an error like 

`subsonic error code 0: find tracks: RIGHT and FULL OUTER JOINs are not currently supported`

in this case it seems like a normal inner join would do the trick? or wdyt?

sentriz (commented):
this is looking great :+1: came across this error though

brian-doherty:
You are of course correct so I've pushed a fix. Question for you: For client who use the directory tree instead of tags, when they star an artist the album entry gets starred. Do you think we should look up the artist and star them as well, and likewise for the artist album when they use tags?

sentriz:
> You are of course correct so I've pushed a fix. Question for you: For client who use the directory tree instead of tags, when they star an artist the album entry gets starred. Do you think we should look up the artist and star them as well, and likewise for the artist album when they use tags?

no i think it's right the way it is now. for the browse by folder spec, the artist object/table doesn't even exist. so i think just staring the artist or album folder is the right way to go

brian-doherty:
OK then I think this is ready to merge.

sentriz:
PS @brian-doherty what do you think about this change? 

since the gonic IDs are always prefixed with their type, I don't think we care about where they come from in the params (id, albumId, artistId etc)

which i think means we can simplify a bit

also just in case you didnt know and you like the diff, you can just save the text as a file then `git apply <file>` and push 

```diff
diff --git a/server/ctrlsubsonic/handlers_by_tags.go b/server/ctrlsubsonic/handlers_by_tags.go
index eb52f67..1bf6c47 100644
--- a/server/ctrlsubsonic/handlers_by_tags.go
+++ b/server/ctrlsubsonic/handlers_by_tags.go
@@ -688,87 +688,54 @@ func (c *Controller) ServeGetSimilarSongsTwo(r *http.Request) *spec.Response {
 	return sub
 }
 
-func (c *Controller) parseStarParams(p params.Params) (*spec.Response, []int, []int, []int) {
-	var albumIDs []int
-	var artistIDs []int
-	var trackIDs []int
+func starIDsOfType(p params.Params, typ specid.IDT) []int {
+	var ids []specid.ID
+	ids = append(ids, p.GetOrIDList("id", nil)...)
+	ids = append(ids, p.GetOrIDList("albumId", nil)...)
+	ids = append(ids, p.GetOrIDList("artistId", nil)...)
 
-	if idl, err := p.GetIDList("id"); err == nil {
-		for _, i := range idl {
-			switch i.Type {
-			case specid.Album:
-				albumIDs = append(albumIDs, i.Value)
-			case specid.Artist:
-				artistIDs = append(artistIDs, i.Value)
-			case specid.Track:
-				trackIDs = append(trackIDs, i.Value)
-			default:
-				return spec.NewError(0, "non-album non-artist non-track id cannot be starred"), nil, nil, nil
-			}
+	var out []int
+	for _, id := range ids {
+		if id.Type != typ {
+			continue
 		}
+		out = append(out, id.Value)
 	}
-
-	if idl, err := p.GetIDList("albumId"); err == nil {
-		for _, i := range idl {
-			switch i.Type {
-			case specid.Album:
-				albumIDs = append(albumIDs, i.Value)
-			default:
-				return spec.NewError(0, "non-album id passed as albumId"), nil, nil, nil
-			}
-		}
-	}
-
-	if idl, err := p.GetIDList("artistId"); err == nil {
-		for _, i := range idl {
-			switch i.Type {
-			case specid.Artist:
-				artistIDs = append(artistIDs, i.Value)
-			default:
-				return spec.NewError(0, "non-artist id passed as artistId"), nil, nil, nil
-			}
-		}
-	}
-
-	return nil, albumIDs, artistIDs, trackIDs
+	return out
 }
 
 func (c *Controller) ServeStar(r *http.Request) *spec.Response {
-	p := r.Context().Value(CtxParams).(params.Params)
-	response, albumIDs, artistIDs, trackIDs := c.parseStarParams(p)
-	if response != nil {
-		return response
-	}
+	params := r.Context().Value(CtxParams).(params.Params)
+	user := r.Context().Value(CtxUser).(*db.User)
 
 	stardate := time.Now()
-	user := r.Context().Value(CtxUser).(*db.User)
-	for _, i := range albumIDs {
+	for _, id := range starIDsOfType(params, specid.Album) {
 		var albumstar db.AlbumStar
-		_ = c.DB.Where("user_id=? AND album_id=?", user.ID, i).First(&albumstar).Error
+		_ = c.DB.Where("user_id=? AND album_id=?", user.ID, id).First(&albumstar).Error
 		albumstar.UserID = user.ID
-		albumstar.AlbumID = i
+		albumstar.AlbumID = id
 		albumstar.StarDate = stardate
 		if err := c.DB.Save(&albumstar).Error; err != nil {
 			return spec.NewError(0, "save album star: %v", err)
 		}
 	}
 
-	for _, i := range artistIDs {
+	for _, id := range starIDsOfType(params, specid.Artist) {
 		var artiststar db.ArtistStar
-		_ = c.DB.Where("user_id=? AND artist_id=?", user.ID, i).First(&artiststar).Error
+		_ = c.DB.Where("user_id=? AND artist_id=?", user.ID, id).First(&artiststar).Error
 		artiststar.UserID = user.ID
-		artiststar.ArtistID = i
+		artiststar.ArtistID = id
 		artiststar.StarDate = stardate
 		if err := c.DB.Save(&artiststar).Error; err != nil {
 			return spec.NewError(0, "save artist star: %v", err)
 		}
 	}
 
-	for _, i := range trackIDs {
+	for _, id := range starIDsOfType(params, specid.Track) {
 		var trackstar db.TrackStar
-		_ = c.DB.Where("user_id=? AND track_id=?", user.ID, i).First(&trackstar).Error
+		_ = c.DB.Where("user_id=? AND track_id=?", user.ID, id).First(&trackstar).Error
 		trackstar.UserID = user.ID
-		trackstar.TrackID = i
+		trackstar.TrackID = id
 		trackstar.StarDate = stardate
 		if err := c.DB.Save(&trackstar).Error; err != nil {
 			return spec.NewError(0, "save track star: %v", err)
@@ -779,40 +746,24 @@ func (c *Controller) ServeStar(r *http.Request) *spec.Response {
 }
 
 func (c *Controller) ServeUnstar(r *http.Request) *spec.Response {
-	p := r.Context().Value(CtxParams).(params.Params)
-	response, albumIDs, artistIDs, trackIDs := c.parseStarParams(p)
-	if response != nil {
-		return response
-	}
-
+	params := r.Context().Value(CtxParams).(params.Params)
 	user := r.Context().Value(CtxUser).(*db.User)
-	for _, i := range albumIDs {
-		var albumstar db.AlbumStar
-		err := c.DB.Where("user_id=? AND album_id=?", user.ID, i).First(&albumstar).Error
-		if err != nil {
-			if err := c.DB.Delete(&albumstar).Error; err != nil {
-				return spec.NewError(0, "delete album star: %v", err)
-			}
+
+	for _, id := range starIDsOfType(params, specid.Album) {
+		if err := c.DB.Where("user_id=? AND album_id=?", user.ID, id).Delete(db.AlbumStar{}).Error; err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+			return spec.NewError(0, "delete album star: %v", err)
 		}
 	}
 
-	for _, i := range artistIDs {
-		var artiststar db.ArtistStar
-		err := c.DB.Where("user_id=? AND artist_id=?", user.ID, i).First(&artiststar).Error
-		if err != nil {
-			if err := c.DB.Delete(&artiststar).Error; err != nil {
-				return spec.NewError(0, "delete artist star: %v", err)
-			}
+	for _, id := range starIDsOfType(params, specid.Artist) {
+		if err := c.DB.Where("user_id=? AND artist_id=?", user.ID, id).Delete(db.ArtistStar{}).Error; err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+			return spec.NewError(0, "delete artist star: %v", err)
 		}
 	}
 
-	for _, i := range trackIDs {
-		var trackstar db.TrackStar
-		err := c.DB.Where("user_id=? AND track_id=?", user.ID, i).First(&trackstar).Error
-		if err != nil {
-			if err := c.DB.Delete(&trackstar).Error; err != nil {
-				return spec.NewError(0, "delete track star: %v", err)
-			}
+	for _, id := range starIDsOfType(params, specid.Track) {
+		if err := c.DB.Where("user_id=? AND track_id=?", user.ID, id).Delete(db.TrackStar{}).Error; err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+			return spec.NewError(0, "delete track star: %v", err)
 		}
 	} 
```

brian-doherty:
Looks good to me. Git is complaining that the patch is corrupt (missing last line) so just go ahead and push it.

sentriz:
shipped! thanks brain :fire: 

also i had another looked and updated a couple small things, hope you dont mind

- store a truncated float for average rating in the db, but a string in the subsonic view (since to separate the data from the view i reckon, maybe in future we'd like to do some queries on the average)
- make some sql keywords uppercase
- make the get starred functions mirror the search functions 
  - which is to say, for getStarred (browse by folder) we dont have artists. so instead an artist is a folder in the root, and an album is a folder whos parent is in the root. this is how it is everywhere else and also i think what subsonic assumes. and these views are very similar to the search views. among others
  - applied the change we already talked about

apologies for not doing discussing these over a review, don't have a lot of time these days. but if have any bones to pick with this please let me know!

and thanks again :)

PR #248, opened 2022-10-27
Various fixes to getAlbumList(2): Added starred, fixed order when fro…

brian-doherty:
…mYear > toYear, added order of alpha by album for byGenre


sentriz:
lookin good, thanks!

PR #252, opened 2022-11-05
Added podcast purger.

brian-doherty:


sentriz:
nice thanks! looks good. i just removed the "(0 disable)" thing, i think it's already implied by the "(optional)"

and also it seems like there's no need to do the purge in a goroutine (potentially we could have 2 purges at once?)

are these changes ok for you?

image

brian-doherty:
OK by me. Thanks!

B

PR #255, opened 2022-11-09
Only purge completed podcast episodes.

brian-doherty:


podcasts/podcasts.go
```diff
@@ -520,7 +520,7 @@ func (p *Podcasts) PurgeOldPodcasts(maxAge time.Duration) error {
 	expDate := time.Now().Add(-maxAge)
 	var episodes []*db.PodcastEpisode
 	err := p.db.
-		Debug().
+		Where("status = ?", "completed").
```
sentriz, line 523: does this also work?

```suggestion
		Where("status = ?", db.PodcastEpisodeStatusCompleted).
```

sentriz:
thanks :+1:

PR #266, opened 2022-12-08
delete the mpv tmp file after creation

### LINKED ISSUE #265: mpv not starting on arm/raspberrypi
from https://github.com/sentriz/gonic/pull/211#issuecomment-1333596043

@marlop352 :wave:

lucat1:
Previously, a temporary path was not only being generated but also touched, therefore the os.Stat check to verify mpv had started was really doing nothing and just succeeding at the first attempt.

Closes #265

In order to test on your machine (a Pi is not needed I believe) you can just add an `mpv` sh script to your `$PATH` (before the actual `mpv`) with just a `sleep 10` inside. This will simulate a somewhat slow process starting (what I guess happens on a Pi). Without this change, the error would have been `failed to connect` as the startup check would always return a false positive. Now it properly hangs for 5s and errors out with `mpv never started`.

If you have the chance @marlop352, give this patch a try on your Pi. I don't currently have one on hand.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -363,6 +363,9 @@ func (s *Server) StartJukebox(mpvExtraArgs []string) (FuncExecute, FuncInterrupt
 			if err != nil {
 				return fmt.Errorf("create tmp sock file: %w", err)
 			}
+			if err := os.Remove(sockFile.Name()); err != nil {
+				return fmt.Errorf("delete tmp sock file: %w", err)
+			}
 			if err := s.jukebox.Start(sockFile.Name(), mpvExtraArgs); err != nil {
 				return fmt.Errorf("start jukebox: %w", err)
 			}


marlop352:
@lucat1 any instructions on how to do so using docker?
I would rather not install anything bare metal on the Pi unless I can't avoid it.

sentriz:
ah i see! thanks makes sense. i went with a slightly different version:

- create a temp dir
- set the sock path to <temp dir>/sock, but don't create it
- let mpv create the sock file in the tmp dir
- cleanup the tmp dir later

the reason being if we create a tmp file, delete it, create it again - there is a very very slim chance that is already exists. where as if we create a tmp dir then go will make sure there's no clash

i did some light testing and seems to work - but let me know if not

thanks!

sentriz:
PS @marlop352 i just kicked off a job to build the nightly image [here](https://github.com/sentriz/gonic/actions/runs/3649243969)

so when it's done you can test out the change with the sentriz/gonic:nightly tag

PR #289, opened 2023-02-12
Add "audiophile" transcode profile

zpuskas:
This diff adds the 192kbps OPUS transcoding profile named opus_audiophile. While 128kbps is already near transparent and good for most use cases, sometimes (or by some people) a bit more is preferred. These are typically for dedicated high quality wireless audio link (A2DP-bluetooth) enabled headphones and Hi-Fi adapters.

The 160-192kbps range is considered to be transparent with very low chance of artifacts (only with very few and specific samples)[1][2].

[1] https://wiki.hydrogenaud.io/index.php?title=Opus
[2] https://ceur-ws.org/Vol-2654/paper25.pdf

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -27,6 +27,7 @@ var UserProfiles = map[string]Profile{
 	"opus_128_car": Opus128RGLoud,
 	"opus_128":     Opus128,
 	"opus_128_rg":  Opus128RG,
+	"opus_192":     Opus192,
 }
 
 // Store as simple strings, since we may let the user provide their own profiles soon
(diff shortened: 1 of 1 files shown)


sentriz:
ok seems reasonable to me. how about name the profile in the same style as we have already? eg we could add 

opus_192_car
opus_192
opus_192_rg

?

zpuskas:
My original idea was that unless using some HiFi type of equipment one would not need this extra bitrate (e.g. in a car), but sure, let's give people more choice. I've updated the PR as per your suggestion.

zpuskas:
As per our IRC discussion, now we only have one Hi-Fi opus profie under the name "opus_192". :)

PR #290, opened 2023-02-13
feat(contrib): improve system related contrib files

zpuskas:
- Make gonic configuration self documenting
- Update systemd files

### DIFF
--- a/contrib/config
+++ b/contrib/config
@@ -1,7 +1,52 @@
-music-path        <path to your music dir>
-podcast-path      <path to your podcasts dir>
-cache-path        <path to cache dir>
-db-path           /var/lib/gonic/gonic.db
-jukebox-enabled   false
-listen-addr       127.0.0.1:4747
-scan-interval     0
+# This is the gonic server system-wide configuration file. For other
+# administrative and per user settings go th the web UI.
+#
+# The strategy used for options in the default config shipped with gonic is to
+# specify options with their default value where possible, and leave optional
+# ones commented. Uncommented options override the default value. Options that
+# are mandatory and need setting before first run have a placeholder for their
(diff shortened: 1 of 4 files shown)


contrib/config
```diff
@@ -1,7 +1,52 @@
...
+
+# Interval (in minutes) to check for new music. Default: don't scan
+#scan-interval               0
+#scan-at-start-enabled       false
+#scan-watcher-enabled        false
+#jukebox-enabled             false
+#jukebox-mpv-extra-args      <extra command line arguments to pass to the jukebox mpv daemon>
+
+# A string or character to split genre tags on for multi-genre support (e.g. ;)
+#genre-split                 "\n"
```
sentriz, line 52: i'm not sure that the config file parser does any string unquoting, or if it's even possible to have a new line in the config like that

```suggestion
#genre-split                 ;
```

contrib/config
```diff
@@ -1,7 +1,52 @@
...
+#http-log                    true
+# URL path prefix to use if behind reverse proxy
+#proxy-prefix                <prefix>
+
+# Secure connection settings. Recommended to set up unless Gonic sits behing an
+# SSL enabled reverse proxy server. Disabled by default.
+#tls-cert                    <path to TLS certificatie>
+#tls-key                     <path to TLS private key>
+
+# Gonic's internal state database location
```
sentriz, line 23:
```suggestion
# gonic's internal state database location
```

sentriz (commented):
thanks! looks great

contrib/config
zpuskas: I've actually found the project's name to be incorrectly capitalized in a few more spots in the configuration file, so I fixed all those too.

Issue #287, opened 2023-02-08
Panic on folder scanning by watcher


mazzz1y:
It happened at the same time when qBittorrent was downloading the music into this folder. 
#### Gonic
```
sentriz/gonic:v0.15.2
GONIC_SCAN_WATCHER_ENABLED=true
```
#### qBittorrent

qBitorrrent has this option enabled(I think that it can be related):
```
Options > Downloads > Append .!qB extension to incomplete files
```
## Logs
```
2023/02/07 12:12:43 processing folder `/music/xxx`
2023/02/07 12:12:53 processing folder `/music/xxx`
2023/02/07 12:13:03 processing folder `/music/xxx`
2023/02/07 12:13:14 processing folder `/music/xxx`
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x8eecec]
goroutine 26 [running]:
	go.senan.xyz/gonic/scanner.(*Scanner).ExecuteWatch(0xc0002ac780)
	/src/scanner/scanner.go:189 +0x3ac
	go.senan.xyz/gonic/server.(*Server).StartScanWatcher.func1()
	/src/server/server.go:354 +0x45
	github.com/oklog/run.(*Group).Run.func1({0xc000413630?, 0xc000413640?})
	/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:38 +0x2f
	created by github.com/oklog/run.(*Group).Run
	/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:37 +0x22a
```

glward:
Seeing the same thing with the watcher enabled -- in my case, the library folder is being synced with Seafile

PR #292, opened 2023-02-17
Fix watcher panic

mazzz1y:
Fixes: https://github.com/sentriz/gonic/issues/287

```golang
fileInfo, err := os.Stat(event.Name)
if err != nil && fileInfo.IsDir()
```

This statement is meaningless because if error is not empty then fileInfo be always empty. It causes panic which is mentioned in the issue.

My solution is just skip cycle if file moved or deleted during the scan.


sentriz:
thanks!

PR #294, opened 2023-02-19
search logic alter

cacko:
change how the search work, currently it just gets the query and use it like it, this PR

* splits the query to words, searches for artist LIKE each word 
* same applied to albums, but includes any found artists in the filter
* same for track but includes if anything is found from above

let's presume you haev all dream theater albums. searching right now for "dream theater space" will return you nothing, after the PR it will return the artist and two songs, lifhting shadow of a dream, space dye vest

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ gonicembed
 *.swp
 .tags*
 *.test
+.envrc
+vendor/
--- a/server/ctrlsubsonic/handlers_by_tags.go
+++ b/server/ctrlsubsonic/handlers_by_tags.go
@@ -209,8 +209,11 @@ func toLikeStrm(vs []string, col string) (string, []sql.NamedArg) {
 	qparts := []string{}
 	qparams := []sql.NamedArg{}
 	for idx, v := range vs {
-		qparts = append(qparts, fmt.Sprintf("%s LIKE @%s_%d OR %s_u_dec LIKE @%s_%d", col, col, idx, col, col, idx))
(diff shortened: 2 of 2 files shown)


sentriz:
hi, i'm having a bit of difficulty understanding what the generated SQL would be here. could you give an example? say I type in "eat static" into the search box - what might the generated SQL be?

cacko:
> hi, i'm having a bit of difficulty understanding what the generated SQL would be here. could you give an example? say I type in "eat static" into the search box - what might the generated SQL be for each table?

artist: name LIKE @name_0 OR name_u_dec LIKE @name_0 OR name LIKE @name_1 OR name_u_dec LIKE @name_1
named params: name_0 = "%eat%", name_1 = "%static%"

same thing goes on for the album and track just the columns names differents.

if artist has returns results, they are added to the album using pretty much same logic but with = instead of like

for the track query, if artist are present they are added, if album are present they are added as above with =

sentriz:
it looks to me like it could be simpler, unless im missing something. 

replacing spaces in the user's query with `%`, performance aside, would have the same effect: eg `eat static` -> `%eat%static%`

https://www.db-fiddle.com/f/cGh6PyENLuWmuFewu9danZ/0

cacko:
dunno, you know best. my point is, if I search for "europe anger" to get band europe, and going downwards to use that artist - no each part to be independent - guess it might be personal preference. feel free to close it.

sentriz:
ahh understood :+1: then I will try this PR locally for a while and see how I like the new way it works. i can also ask on the IRC channel if anyone else has any preference on this. closing for now, thanks!

PR #295, opened 2023-02-20
lastfm artist image h4ck

cacko:


cacko:
it is just a working concept on my work, I hope somebody can pick it and code it properly in go. my go knowledge is around 0, just common sense logic from other languages I use.

thanks

cacko:
Screenshot 2023-02-20 at 21 21 35
Screenshot 2023-02-20 at 21 21 21

here are some proofs, just by using getArtistInfo2 call and that lastfm mod

sentriz:
thanks! nice (and cheeky) idea. though i decided to use the go html and cascadia packages instead of opengraph - since they were already dependencies of the project

PR #314, opened 2023-05-07
Bugfix: Add playlists directory to volumes

gzurowski:
What?
=====
Add the `/playlists` directory to the list of volumes, so the server doesn't exit when validating the playlist path on startup.

Why?
=====
With the current behavior, the server process exits when no playlist directory is found, see: https://github.com/sentriz/gonic/blob/7dc9575e52538cf4fd39715193b5c760e1cc477b/cmd/gonic/gonic.go#L104

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ COPY --from=builder \
 COPY --from=builder \
     /src/gonic \
     /bin/
-VOLUME ["/cache", "/data", "/music", "/podcasts"]
+VOLUME ["/cache", "/data", "/music", "/playlists", "/podcasts"]
 EXPOSE 80
 ENV TZ ""
 ENV GONIC_DB_PATH /data/gonic.db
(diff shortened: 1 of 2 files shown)


sentriz:
thanks 👍

sentriz:
wait, I just realised that probably exiting like that is behaviour that we want for users who are upgrading from the latest release (where playlists are in the db, not filesystem)

 since we have a migration that copies from the db to GONIC_PLAYLISTS_PATH, if that volume is created automatically without the user mounting it, they loose their playlists

gzurowski:
> wait, I just realised that probably exiting like that is behaviour that we want for users who are upgrading from the latest release (where playlists are in the db, not filesystem)
> 
> since we have a migration that copies from the db to GONIC_PLAYLISTS_PATH, if that volume is created automatically without the user mounting it, they loose their playlists
> 
> so by exiting, we force the user to explicitly mount a non ephemeral playlists dir so there's hopefully no data loss

Definitely a valid point, I wasn't aware of that. But the current behavior is also not ideal for new users as it enforces to specify a non-ephemeral playlist directory even if not such functionality is not used. Is it possible to apply such check (i.e., for an existing directory) to the trigger of the database migration?

PR #317, opened 2023-05-11
Add unit tests for ListenBrainz scrobbler

gzurowski:
Add some initial tests for the ListenBrainz scrobbler.


sentriz:
hi @gzurowski this looks quite nice :+1: 

how do you feel about these changes? basically just using the go:embed directive to get the test data (since i found the t.Name() to file lookup a bit confusing, and seems like we'd be limited to one bit of test data per test)

<details>
<summary>👇 diff</summary>

```diff
diff --git a/scrobble/listenbrainz/listenbrainz_test.go b/scrobble/listenbrainz/listenbrainz_test.go
index d2b1141..3ed0947 100644
--- a/scrobble/listenbrainz/listenbrainz_test.go
+++ b/scrobble/listenbrainz/listenbrainz_test.go
@@ -3,14 +3,11 @@ package listenbrainz
 import (
 	"context"
 	"crypto/tls"
+	_ "embed"
 	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
-	"os"
-	"path"
-	"regexp"
-	"strings"
 	"testing"
 	"time"
 
@@ -34,6 +31,9 @@ func httpClientMock(handler http.Handler) (http.Client, func()) {
 	return client, server.Close
 }
 
+//go:embed testdata/submit_response.json
+var submitResponse string
+
 func TestScrobble(t *testing.T) {
 	t.Parallel()
 	assert := assert.New(t)
@@ -46,7 +46,7 @@ func TestScrobble(t *testing.T) {
 		assert.Equal("Token token1", r.Header.Get("Authorization"))
 		bodyBytes, err := io.ReadAll(r.Body)
 		assert.NoError(err)
-		assert.JSONEq(getTestData(t), string(bodyBytes))
+		assert.JSONEq(submitResponse, string(bodyBytes))
 
 		w.WriteHeader(http.StatusOK)
 		w.Write([]byte(`{"accepted": 1}`))
@@ -74,7 +74,7 @@ func TestScrobble(t *testing.T) {
 	assert.NoError(err)
 }
 
-func TestScrobble_unauthorized(t *testing.T) {
+func TestScrobbleUnauthorized(t *testing.T) {
 	t.Parallel()
 	assert := assert.New(t)
 
@@ -111,7 +111,7 @@ func TestScrobble_unauthorized(t *testing.T) {
 	assert.ErrorIs(err, ErrListenBrainz)
 }
 
-func TestScrobble_serverError(t *testing.T) {
+func TestScrobbleServerError(t *testing.T) {
 	t.Parallel()
 	assert := assert.New(t)
 
@@ -146,24 +146,3 @@ func TestScrobble_serverError(t *testing.T) {
 	// assert
 	assert.ErrorIs(err, ErrListenBrainz)
 }
-
-func getTestData(t *testing.T) string {
-	t.Helper()
-
-	dataPath := getTestDataPath(t.Name())
-	bytes, err := os.ReadFile(dataPath)
-	if err != nil {
-		t.Fatalf("failed to read test data: %v", err)
-	}
-	return string(bytes)
-}
-
-var testCamelExpr = regexp.MustCompile("([a-z0-9])([A-Z])")
-
-func getTestDataPath(test string) string {
-	// convert test name to query case path
-	snake := testCamelExpr.ReplaceAllString(test, "${1}_${2}")
-	lower := strings.ToLower(snake)
-	relPath := strings.ReplaceAll(lower, "/", "_") + ".json"
-	return path.Join("testdata", relPath)
-}
diff --git a/scrobble/listenbrainz/testdata/test_scrobble.json b/scrobble/listenbrainz/testdata/submit_response.json
similarity index 100%
rename from scrobble/listenbrainz/testdata/test_scrobble.json
rename to scrobble/listenbrainz/testdata/submit_response.json
```
</details>

and regarding using `assert` vs `is`. i really quite like the ergonomics of `is` and how the comments next to the assertion show up in the output. and it's surface area is so small which is also nice

for JSONEq we could just have `is.True(jsonEq(a, b)) // submit json is equal`

but i don't mind too much either way. if we go for assert than we probably want to make an issue to convert the existing usage of `is` to `assert`

gzurowski:
> how do you feel about these changes? basically just using the go:embed directive to get the test data (since i found the t.Name() to file lookup a bit confusing, and seems like we'd be limited to one bit of test data per test)

Thanks for the suggestion and the diff. Really like the use of the `go:embed` directive with less boilerplate and probably better maintainability.

> but i don't mind too much either way. if we go for assert than we probably want to make an issue to convert the existing usage of `is` to `assert`

Testify is well-maintained and has a wide range of assertion functions which make it more expressive and probably easier on the eye (e.g. `assert.IsError(err, TargetErr)` vs `is.True(errors.Is(err, targetErr)`, etc.). I'd be happy to log a corresponding issue and work on the conversion.

gzurowski:
> there is an is.NoErr() , but if you're happy to covert existing usage of is to assert then sounds good 👍

I've created an initial PR: #319.

sentriz:
nice thanks!

PR #319, opened 2023-05-13
Use Testify for all tests

gzurowski:


sentriz (commented):
ah apologies i just noticed, to keep the old functionality, i think it should be mapped like


is.New() -> require
is.NewRelaxed() -> assert

require is from the same package "github.com/stretchr/testify/require"

gzurowski:
> ah apologies i just noticed, to keep the old functionality, i think it should be mapped like
> 
> is.New() -> require is.NewRelaxed() -> assert
> 
> require is from the same package "github.com/stretchr/testify/require"

Thanks for pointing this out. I've implemented the corresponding changes.

sentriz:
cool thanks!

PR #323, opened 2023-05-16
Refactor Last.FM client

gzurowski:
- Separate client and scrobbler.
- Move model into own file and fix attribute for playcount.
- Inject HTTP client into Last.FM client, so it's easier to write tests.
- Update credentials once changed instead of continuously reading them from database.
- Add an initial unit test for`artist.getInfo`


server/ctrladmin/handlers.go
```diff
@@ -366,6 +357,7 @@ func (c *Controller) ServeUpdateLastFMAPIKeyDo(r *http.Request) *Response {
 	if err := c.DB.SetSetting("lastfm_secret", secret); err != nil {
 		return &Response{redirect: r.Referer(), flashW: []string{fmt.Sprintf("couldn't set secret: %v", err)}}
 	}
+	c.lastfmClient.UpdateAPIKey(apiKey, secret)
```
sentriz, line 360: i'm not sure about this concept of keeping state like api key/secret in memory as part of a client object. it feels like they could go out of sync easily. is it an OOP pattern? i'm not so familiar with OOP stuff

gzurowski: It is a common approach to inject configuration into a target instance, but I agree that the mechanism to keep it up-to-date must be more robust. I will close this PR in favor of new ones to separate testing improvements and the management of configuration (so that we don't need to re-query the database on every operation).

PR #325, opened 2023-05-19
Update linter

gzurowski:
- Update Github [golangci-lint-action](https://github.com/golangci/golangci-lint-action) to v3.
- Update [golangci-lint](https://github.com/golangci/golangci-lint) to v1.52.2.
- Fix all linter issues (remove/_ unused arguments, stop using reserved keywords).

### DIFF
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -33,10 +33,9 @@ jobs:
           sudo apt update -qq
           sudo apt install -y -qq build-essential git sqlite libtag1-dev ffmpeg mpv zlib1g-dev
       - name: Lint
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.50.1
-          skip-go-installation: true
+          version: v1.52.2
           args: --timeout=5m
       - name: Test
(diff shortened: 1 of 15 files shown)


.github/workflows/nightly-release.yaml
```diff
@@ -33,10 +33,9 @@ jobs:
           sudo apt update -qq
           sudo apt install -y -qq build-essential git sqlite libtag1-dev ffmpeg mpv zlib1g-dev
       - name: Lint
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.50.1
-          skip-go-installation: true
+          version: v1.52.2
```
sentriz, line 38: why do we get rid if skip-go-installation? iirc it was true because we have the setup go action above

gzurowski: The option has been removed in the new version of `golangci-lint-action`, see: https://github.com/golangci/golangci-lint-action#compatibility:

> v3.0.0+ requires explicit setup-go installation step prior to using this action: uses: actions/setup-go@v3. The skip-go-installation option has been removed.

sentriz:
cool thanks!

PR #327, opened 2023-05-19
chore: Update to Go 1.20

gzurowski:
- Update Go to 1.20
- Update all dependencies
- Update Dockerfiles to use base images with Go 1.20


gzurowski:
> i don't think we're using any go 1.20 specific features, while making it a bit more difficult for users to run gonic baremetal (eg the go version available from debian stable)

That's a fair point. I mainly wanted to keep the code up-to-date in respect to the language version and its dependencies, not being aware of the requirement for custom builds on different operating systems.

Do you think providing rpm/deb/etc. packages would relax the requirement of staying at a particular Go version going forward?

sentriz:
honestly im not sure how packages are built for different distros. though there is a packaging section in the readme, maybe some of them are Debian like

PR #329, opened 2023-05-20
Add initial Last.FM tests

gzurowski:
- Refactor package: separate model, scrobbler and client implementation.
- Fix playcount attribute in existing model.
- Add an initial test for `ArtistGetInfo`.


sentriz:
cool thanks :+1:

PR #331, opened 2023-05-31
Changed frequent album list to use total time played per album instea…

### LINKED ISSUE #285: Should "most played" list divide by track count?
The "most played" list tends to overweight really long albums (box sets especially) because it is sorted by track plays and they have so many track. I submit as a question -- should it divide by track count instead, to give a notion of which _albums_ you play the most?

brian-doherty:
…d of play count. Closes #285


sentriz:
not near computer atm but seem to have a migration error

sentriz:
do we need an AutoMigrate() at the top of the migration body?

brian-doherty:
Huh, i didn't have any problem in my local dev. Am i missing an ALTER TABLE to add the column? I'd try something but I'm on the road.

sentriz:
fixed now, i think

PR #339, opened 2023-07-17
Need to add a safeFilename call going into findUniqueEpisodeName.

brian-doherty:
This was causing me problems with the Equity podcast, which uses GUIDs as filenames.


sentriz:
thanks!

Issue #44, opened 2020-03-11
external user authentification


unclesamwk:
Hey,

subsonic supports authentication against a ldap server. This feature would be nice. Additionally the oauth support.

thanks
sam

sentriz:
this would be a cool feature alright. im wondering about how much complexity it would add to the codebase though

unclesamwk:
Very nice LDAP Server 
https://github.com/glauth/glauth

helmut72:
LDAP authentication would be great. All of my other self hosted services use LDAP to have a single point of user administration.

helmut72:
Maybe some http header based auth with automatic user creation is easier to implement? Then everyone can choose the prefered auth system (ldap, openid connect, saml...). I use http header based auth with FreshRSS and it works really great. To connect the any apps, the user must login and create an "app password" of course.

sentriz:
looking at this again, I'm not sure its possible given the subsonic API's auth mechanism using query params. if anyone has any other ideas please reopen 👍

BetaPictoris:
I'm not sure about how it was implemented, but Airsonic-Advanced has LDAP authentication.

PR #346, opened 2023-08-06
[WIP] LDAP Authentication

BetaPictoris:
## To-do

- [ ] LDAP configuration page.
  - ~~Configuration on the admin page~~ (`/ldap`)
  - ~~Saving the configuration to the database~~ (`/ldap_do`)
  - Reading the current configuration and setting the values on `/ldap`
- [ ] Connect to the LDAP server.
- [ ] Log in using an LDAP account.
  - Check the user's credentials against the LDAP server.
  - Automatically create a user account if there is a specific LDAP group (ie. `media`).

## Limitations

- Currently the bind user's password is stored unencrypted, this seems to be a limitation of [`go-ldap`](https://github.com/go-ldap/ldap).
- There may be users who the LDAP server believes should be an administrator (ie. of group `media_admin`), currently this implementation has no filter for detecting an admin user. 

---

Fixes: #44.

### DIFF
--- a/server/ctrladmin/handlers.go
+++ b/server/ctrladmin/handlers.go
@@ -168,7 +168,7 @@ func (c *Controller) ServeLDAPConfig(r *http.Request) *Response {
 
 func (c *Controller) ServeLDAPConfigDo(r *http.Request) *Response {
 	bindUID := r.FormValue("bind_user")
-	bindPSWD := r.FormValue("bind_user_password")
+	bindPSWD := r.FormValue("bind_user_password") // TODO: Secure this better.
 	fqdn := r.FormValue("fqdn")
 	port := r.FormValue("port")
 	tls := r.FormValue("tls")
--- a/server/ctrladmin/handlers.go
+++ b/server/ctrladmin/handlers.go
@@ -157,6 +157,15 @@ func (c *Controller) ServeUnlinkListenBrainzDo(r *http.Request) *Response {
 	return &Response{redirect: "/admin/home"}
 }
 
+func (c *Controller) ServeLDAPConfig(r *http.Request) *Response {
+	data := &templateData{}
+
+	return &Response{
+		template: "ldap.tmpl",
+		data: data,
+	}
+}
+
 func (c *Controller) ServeChangeUsername(r *http.Request) *Response {
 	user, err := selectedUserIfAdmin(c, r)
 	if err != nil {
--- a/server/ctrladmin/handlers.go
+++ b/server/ctrladmin/handlers.go
@@ -166,6 +166,34 @@ func (c *Controller) ServeLDAPConfig(r *http.Request) *Response {
 	}
 }
 
+func (c *Controller) ServeLDAPConfigDo(r *http.Request) *Response {
+	bindUID := r.FormValue("bind_user")
+	bindPSWD := r.FormValue("bind_user_password")
+	fqdn := r.FormValue("fqdn")
+	port := r.FormValue("port")
+	tls := r.FormValue("tls")
+	baseDN := r.FormValue("base_dn")
+	filter := r.FormValue("filter")
+	
+	if bindUID == "" || bindPSWD == "" || fqdn == "" || port == "" || port == "" || baseDN == "" || filter == "" {
+		return &Response{
+			redirect: r.Referer(),
+			flashW:   []string{"please provide a bind username and password, a FQDN, port, base CN, and filter"},
+		}
+	}
+	
+	c.DB.SetSetting("ldap_bind_user", bindUID)
+	c.DB.SetSetting("ldap_bind_user_password", bindPSWD)
+	
+	c.DB.SetSetting("ldap_fqdn", fqdn)
+	c.DB.SetSetting("ldap_port", port)
+	c.DB.SetSetting("ldap_tls", tls)
+	c.DB.SetSetting("ldap_base_dn", baseDN)
+	c.DB.SetSetting("ldap_filter", filter)
+
+	return &Response{redirect: "/admin/home"}
+}
+
 func (c *Controller) ServeChangeUsername(r *http.Request) *Response {
 	user, err := selectedUserIfAdmin(c, r)
 	if err != nil {
--- a/server/ctrladmin/adminui/pages/home.tmpl
+++ b/server/ctrladmin/adminui/pages/home.tmpl
@@ -35,7 +35,9 @@
         {{ end }}
     {{ end }}
     {{ if .User.IsAdmin }}
-        <div class="col-span-full">{{ component "link" (props . "To" (path "/admin/create_user")) }}create new{{ end }}</div>
+				<div class="col-span-full">{{ component "link" (props . "To" (path 
+				"/admin/create_user")) }}create new{{ end }} {{component "link" 
+				(props . "To" (path "/admin/ldap")) }}LDAP{{ end }}</div>
     {{ end }}
 </div>
 {{ end }}
--- a/server/ctrladmin/adminui/pages/ldap.tmpl
+++ b/server/ctrladmin/adminui/pages/ldap.tmpl
@@ -35,13 +35,16 @@
 					value="389"
 				>
 
-				
-
-				<label>
+				<label >
 					<input
 						type="checkbox"
 						id="tls"
 						name="tls"
+						value="TLS"
+						style="
+						all: revert !important;
+						width: 20px !important;
+						height: 20px !important;"
 					>
 					TLS
 				</label>
--- a/server/ctrladmin/adminui/pages/ldap.tmpl
+++ b/server/ctrladmin/adminui/pages/ldap.tmpl
@@ -1,0 +1,69 @@
+{{ component "layout" . }}
+{{ component "layout_user" . }}
+
+{{ component "block" (props .
+    "Icon" "user"
+		"Name" "configuring LDAP"
+) }}
+		<form class="flex flex-col gap-2 items-end" action="{{ path 
+		"/admin/ldap_do" }}" method="post">
+				<input
+					type="username"
+					id="bind_user_uid"
+					name="bind_user"
+					placeholder="bind username"
+				>
+				<input
+					type="password"
+					id="bind_user_password"
+					name="bind_user_password"
+					placeholder="bind password"
+				>
+				
+				<input
+					type="text"
+					id="fqdn"
+					name="fqdn"
+					placeholder="LDAP server"
+				>
+				
+				<input
+					type="number"
+					id="port"
+					name="port"
+					placeholder="port"
+					value="389"
+				>
+
+				
+
+				<label>
+					<input
+						type="checkbox"
+						id="tls"
+						name="tls"
+					>
+					TLS
+				</label>
+
+				<input
+					type="text"
+					id="base_cn"
+					name="base_cn"
+					placeholder="Base CN"
+				>
+				
+				<input
+					type="text"
+					id="filter"
+					name="filter"
+					placeholder="filter"
+				>
+				
+
+				<input type="submit" value="update">
+    </form>
+{{ end }}
+
+{{ end }}
+{{ end }}
--- a/server/ctrladmin/adminui/pages/ldap.tmpl
+++ b/server/ctrladmin/adminui/pages/ldap.tmpl
@@ -9,7 +9,7 @@
 		"/admin/ldap_do" }}" method="post">
 				<input
 					type="username"
-					id="bind_user_uid"
+					id="bind_user"
 					name="bind_user"
 					placeholder="bind username"
 				>
@@ -51,9 +51,9 @@
 
 				<input
 					type="text"
-					id="base_cn"
-					name="base_cn"
-					placeholder="Base CN"
+					id="base_dn"
+					name="base_dn"
+					placeholder="Base DN"
 				>
 				
 				<input
--- a/server/server.go
+++ b/server/server.go
@@ -173,6 +173,7 @@ func setupAdmin(r *mux.Router, ctrl *ctrladmin.Controller) {
 	routUser.Use(ctrl.WithUserSession)
 	routUser.Handle("/logout", ctrl.HR(ctrl.ServeLogout)) // "raw" handler, updates session
 	routUser.Handle("/home", ctrl.H(ctrl.ServeHome))
+	routUser.Handle("/ldap", ctrl.H(ctrl.ServeLDAPConfig))
 	routUser.Handle("/change_username", ctrl.H(ctrl.ServeChangeUsername))
 	routUser.Handle("/change_username_do", ctrl.H(ctrl.ServeChangeUsernameDo))
 	routUser.Handle("/change_password", ctrl.H(ctrl.ServeChangePassword))
--- a/server/server.go
+++ b/server/server.go
@@ -174,6 +174,7 @@ func setupAdmin(r *mux.Router, ctrl *ctrladmin.Controller) {
 	routUser.Handle("/logout", ctrl.HR(ctrl.ServeLogout)) // "raw" handler, updates session
 	routUser.Handle("/home", ctrl.H(ctrl.ServeHome))
 	routUser.Handle("/ldap", ctrl.H(ctrl.ServeLDAPConfig))
+	routUser.Handle("/ldap_do", ctrl.H(ctrl.ServeLDAPConfigDo))
 	routUser.Handle("/change_username", ctrl.H(ctrl.ServeChangeUsername))
 	routUser.Handle("/change_username_do", ctrl.H(ctrl.ServeChangeUsernameDo))
 	routUser.Handle("/change_password", ctrl.H(ctrl.ServeChangePassword))


sentriz:
thanks for the effort here :+1: 

it also may take some time for me to review since i don't know anything about LDAP. and i don't think i should blindy accept something i don't know about (no offense :grin:) i will do some reading or get a second pair of eyeballs on it

there seems to be some lint errors btw

BetaPictoris:
> thanks for the effort here 👍
> 
> it also may take some time for me to review since i don't know anything about LDAP. and i don't think i should blindy accept something i don't know about (no offense 😁) i will do some reading or get a second pair of eyeballs on it
> 
> there seems to be some lint errors btw

I totally understand about randomly merging a pull request, specifically when it comes to authentication. Thanks for the heads up!

nitnelave (commented):
Hey! I'm the author of LLDAP, an LDAP server. I saw that @sentriz was not comfortable with LDAP, so I went ahead and added my thoughts on the LDAP parts of the PR. I left a few comments, but nothing major. The LDAP usage here seems straightforward and correct.
Looking forward to LDAP support in gonic!

server/ctrlsubsonic/middleware.go
```diff
@@ -68,18 +72,132 @@ func (c *Controller) WithUser(next http.Handler) http.Handler {
...
+				defer l.Close()
+				
+				// After we have a connection, let's try binding
+				_, err = l.SimpleBind(&ldap.SimpleBindRequest{
+					Username: fmt.Sprintf("uid=%s,%s", bindUID, baseDN),
+					Password: bindPWD,
+				})
+				if err != nil {
+					log.Println("Failed to bind to LDAP:", err)
+					_ = writeResp(w, r, spec.NewError(40, "invalid username `%s`", username))
```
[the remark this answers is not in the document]
BetaPictoris: I've updated the error message to match the [Subsonic API](http://www.subsonic.org/pages/api.jsp).

server/ctrlsubsonic/middleware.go
```diff
@@ -68,18 +72,132 @@ func (c *Controller) WithUser(next http.Handler) http.Handler {
...
+					[]string{"dn"},
+					nil,
+				)
+
+				result, err := l.Search(searchReq)
+				if err != nil {
+        	log.Println("failed to query LDAP:", err)
+				}
+
+  		  if len(result.Entries) > 0 {
```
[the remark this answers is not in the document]
BetaPictoris: I've added a check for this.

server/ctrlsubsonic/middleware.go
```diff
@@ -68,18 +72,132 @@ func (c *Controller) WithUser(next http.Handler) http.Handler {
...
+					user := db.User{
+						Name: username,
+						Password: "", // no password because we want auth to fail.
```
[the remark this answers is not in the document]
BetaPictoris: You can't login without a password, or salt and token, being specified.

BetaPictoris:
As a heads up a99d63ab12838a1b16771abd6cf200d82b2c0b51 removed `not null` from the database model for `User.Password`, so this pull request isn't exactly ready yet - as that is going to require a migration.

server/ctrlsubsonic/middleware.go
```diff
@@ -68,18 +72,132 @@ func (c *Controller) WithUser(next http.Handler) http.Handler {
...
+				// Now, we can try to connect to the LDAP server.
+				l, err := ldap.DialURL(fmt.Sprintf("%s://%s:%s", protocol, ldapFQDN, ldapPort))
+				if err != nil {
+					// Warn the server and return a generic error.
+					log.Println("Failed to connect to LDAP server", err)
+					
+					_ = writeResp(w, r, spec.NewError(0, "Failed to connect to LDAP server."))
+					return
+				}
+				defer l.Close()
```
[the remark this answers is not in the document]
BetaPictoris: Done in 667cf731fd2c6ff8e7142d0f5e1b74c675a7df45. 

PS: Thanks for the review, and making LLDAP!

sentriz:
thank you @nitnelave and @BetaPictoris for the updates and review :+1: @BetaPictoris i will check it out shortly. would you mind rebasing in the meantime? thanks!

nitnelave:
@BetaPictoris there seems to be more conflicts that crept up in the meantime. Can you have a look?

BetaPictoris:
Hey y'all, I've fixed conflicts, sorry that it took a while!

server/ctrlsubsonic/ctrl.go
```diff
@@ -218,43 +220,159 @@ func withRequiredParams(next http.Handler) http.Handler {
...
-			passwordAuth := token == "" && salt == ""
-			tokenAuth := password == ""
-			if tokenAuth == passwordAuth {
-				_ = writeResp(w, r, spec.NewError(10,
-					"please provide `t` and `s`, or just `p`"))
-				return
-			}
-			user := dbc.GetUserByName(username)
-			if user == nil {
+func (c *Controller) WithUser(next http.Handler) http.Handler {
```
sentriz, line 223: looks like this function signature has changed, it used to be `unc withUser(dbc *db.DB) handlerutil.Middleware {` and still can be by the looks of things

server/ctrlsubsonic/ctrl.go
```diff
@@ -218,43 +220,159 @@ func withRequiredParams(next http.Handler) http.Handler {
...
+		passwordAuth := token == "" && salt == ""
+		tokenAuth := password == ""
+		if tokenAuth == passwordAuth {
+			_ = writeResp(w, r, spec.NewError(10,
+				"please provide `t` and `s`, or just `p`"))
+			return
+		}
+		user := c.dbc.GetUserByName(username)
+
+    newLDAPUser := false
```
sentriz, line 241: im finding the control flow quite hard to reason about here, can we do something like

- collect passwordAuth and tokenAuth
- if user != nil { check then and error or ServeHTTP; return }
- otherwise we do LDAP stuff

sentriz (commented):
hey sorry this look so long, only having a look now 🫢

add some comments on the code there i also have a couple general questions

- it looks like for every single question to the server, we go though the withUser middleware and create a new connection to the LDAP server, do the bind, search, etc stuff. seems like a lot of work. is there no token exchange or something so that the next time the user visists they can skip all that? 
- it would be nice if i could use LDAP from the get go. right now it seems like an admin must first visit the dashboard via normal means, then setup LDAP, future visits? seems like it defaults the purpose for single-user servers 
- it's a little tedious to keep calling database methods to get LDAP settings. should they intead be environment variables? maybe it makes sense since these are global settings that we don't want to change at runtime i imagine? also perhaps helps with the above point?
- looks also like go-ldap/ldap is being used but the latest version is /v3

BetaPictoris:
> hey sorry this look so long, only having a look now 🫢

All good, thanks for the comments! 


> it looks like for every single question to the server, we go though the withUser middleware and create a new connection to the LDAP server, do the bind, search, etc stuff. seems like a lot of work. is there no token exchange or something so that the next time the user visists they can skip all that?

There is no token exchange (at least not that I know of in standard LDAP) but we could connect and bind at the server start, or at the first use of LDAP.  This would remove that extra workload, but at the cost of having an open LDAP connection sitting around.


> it would be nice if i could use LDAP from the get go. right now it seems like an admin must first visit the dashboard via normal means, then setup LDAP, future visits? seems like it defaults the purpose for single-user servers

There shouldn't be any future visits to the LDAP page, unless the configuration changes.  These fields could be moved/copied into an environment variables if it would better fit the rest of the setup, in your regards. 


> whats the reason the LDAP users cant have dashboards? how would they change their transcode settings or add internet ratio stations for example? or setup scrobbling?

This could be added, I just haven't gotten around to it, but at the moment `withUser` and `ServeLoginDo` are entirely separate.  I could likely move the LDAP feature out of `withUser` and both `withUser` and `ServeLoginDo` could point to a single function.


> looks also like go-ldap/ldap is being used but the latest version is /v3

I'll update go-ldap.

betapictoris:
@sentriz 

> whats the reason the LDAP users cant have dashboards? how would they change their transcode settings or add internet ratio stations for example? or setup scrobbling?

LDAP users now have dashboards (f6ecd6d39b0a46bd2f8ae590021533ccb0071277).

server/ctrlsubsonic/ctrl.go
```diff
@@ -93,7 +94,7 @@ func New(dbc *db.DB, scannr *scanner.Scanner, musicPaths []MusicPath, podcastsPa
 	chain := handlerutil.Chain(
 		withParams,
 		withRequiredParams,
-		withUser(dbc),
+		c.withUser,
```
sentriz, line 97: we don't need to make this a method of controller any more

just `withUser(dbc)` like it was before, and withUser can look like 

```
func withUser(dbc *db.DB) handlerutil.Middleware {
    return func(next http.Handler) http.Handler {
        return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```

(see the previous implementation)

server/ctrlsubsonic/ctrl.go
```diff
@@ -223,43 +224,47 @@ func withRequiredParams(next http.Handler) http.Handler {
...
+		}
+		user := c.dbc.GetUserByName(username)
+
+		var credsOk bool
+		if tokenAuth && user != nil {
+			credsOk = checkCredsToken(user.Password, token, salt)
+		} else if user != nil {
+			credsOk = checkCredsBasic(user.Password, password)
+		}
+		if !credsOk {
```
sentriz, line 251: not sure i like the idea of "if user/pass fails, try LDAP" 

seems like it would be better to have

```
if ldap is enabled {
   do ldap
   return
}
do normal user name password stuff
```

this would require that LDAP is configured from the very start (see other comment)

server/ctrladmin/handlers_raw.go
```diff
@@ -18,10 +19,18 @@ func (c *Controller) ServeLoginDo(w http.ResponseWriter, r *http.Request) {
 	}
 	user := c.dbc.GetUserByName(username)
 	if user == nil || password != user.Password {
-		sessAddFlashW(session, []string{"invalid username / password"})
-		sessLogSave(session, w, r)
-		http.Redirect(w, r, r.Referer(), http.StatusSeeOther)
-		return
+		// Because internal authentication failed, we can now try to use LDAP, if
```
sentriz, line 22: same as other comment (ill just paste it since i can't link it think)

-------

not sure i like the idea of "if user/pass fails, try LDAP" 

seems like it would be better to have

```
if ldap is enabled {
   do ldap
   return
}
do normal user name password stuff
```

this would require that LDAP is configured from the very start (see other comment)

sentriz:
sorry for the delay! and thanks for updating, it looks a lot better now i think. have one question though, 

when using this over the subsonic API, we are using the password param to do an BindRequest. how does that work, and which clients support this?

nitnelave:
Unless I'm mistaken, it's invisible to the client. They send their credentials to the server, and the server decides how to check the password: it can be built-in user management, or forwarding the credentials to an LDAP server for validation: a successful bind means that the credentials are valid.
The client API doesn't change.

nitnelave:
Actually, they wouldn't even _have_ a subsonic password: you can dynamically create a subsonic user after a successful bind. "I didn't know about this user, but the LDAP server tells me they exist and have access to subsonic, so I'll create an account for them"

Usually, the flow goes like:
- user sends their credentials to subsonic
- subsonic connects to the LDAP server as admin, and checks that the user is allowed to access subsonic (that can be either "user exists" or often "user belongs to the subsonic_users group").
- if the user is allowed access, subsonic connects to the LDAP server using the user's credentials.
- if that succeeds, the user is logged in. If necessary, create a subsonic account to store preferences and so on.

When querying the LDAP server (usually as the admin to save one query), you can also fetch some more details from LDAP, such as the user's email and/or display name.

sentriz:
> Actually, they wouldn't even _have_ a subsonic password: you can dynamically create a subsonic user after a successful bind. "I didn't know about this user, but the LDAP server tells me they exist and have access to subsonic, so I'll create an account for them"
> 
> Usually, the flow goes like:
> 
>     * user sends their credentials to subsonic
> 
>     * subsonic connects to the LDAP server as admin, and checks that the user is allowed to access subsonic (that can be either "user exists" or often "user belongs to the subsonic_users group").
> 
>     * if the user is allowed access, subsonic connects to the LDAP server using the user's credentials.
> 
>     * if that succeeds, the user is logged in. If necessary, create a subsonic account to store preferences and so on.
> 
> 
> When querying the LDAP server (usually as the admin to save one query), you can also fetch some more details from LDAP, such as the user's email and/or display name.

thanks for the info :+1: also, and from gonic's perspective would it be standard practice to hit the LDAP for every http request is receives? seems like a lot of overhead, esp considering the case of browsing a list of albums from a browser. that could be 100s of request to gonic like getCoverArt.view?id=x&p=xxx, which is in turning hitting the LDAP server for each request

nitnelave:
No, logging a user in can be an expensive operation (think multiple rounds of cryptographic hashing if you want something secure). Repeated attempts to log in can fairly easily bring an authentication server to its knees if it's not protected.

Usually you'd call the LDAP server once when logging in, then remember the login. Either through a session token, some JWT or similar, you'd get a short-to-medium term authorization (e.g. 1h, 24h, until 20min of inactivity, until explicit logout, etc). Note that the user's access to the server (gonic) can change at any time and you won't be notified of that; you'll only see that when they next try to log in. As a result, keeping the session until the user explicitly logs out is not recommended (but could be offered as an option for those who want it, along with a way to forcefully invalidate the user's active sessions).

server/ctrlsubsonic/ctrl.go
betapictoris: So, just double checking, you want only LDAP or internal authentication at any given point, right?  And if LDAP authentication is enabled, should the default admin user still exist -- or should there be a separate admin filter (which would allow for more than one admin user) that is used to validate users?

sentriz: hmm i'm not sure honestly. i do like the idea of multiple admin roles, and an admin filter. though gonic will always create an admin user when it runs the first db migrations. so that's probably hard to avoid

betapictoris:
> No, logging a user in can be an expensive operation (think multiple rounds of cryptographic hashing if you want something secure). Repeated attempts to log in can fairly easily bring an authentication server to its knees if it's not protected.

The best why I can think of doing this is syncing passwords from the LDAP server (ie. checking local accounts, querying the LDAP server if it fails, then create the user if it doesn't exist, update the password if it's out-of-date).  This does have a few caveats (as for every failed login there would be a request to the LDAP server, which could be avoided with some sort of rate limit -- but that leads to out of date credentials).  Other than this, I'm not sure if there is any amazing way of implementing this, as far as I know (standard) LDAP doesn't have any token exchange, nor does OpenSubsonic.

nitnelave:
Getting the password (hashes) from LDAP is not recommended, both in terms of security (you don't want to spread them everywhere and increase the attack surface) and compatibility (not every server hashes them in a compatible way). For instance, LLDAP doesn't expose the password hashes because it's doing a zero-knowledge proof that basically can't be replicated in gonic.

How does normal auth work in gonic? The credentials are sent with every request? You can implement a server-side auth caching method, i.e. "these credentials were authorized by the LDAP server at time T, so they'll be valid until T+24h before we recheck them"

betapictoris:
> Getting the password (hashes) from LDAP is not recommended, both in terms of security (you don't want to spread them everywhere and increase the attack surface) and compatibility (not every server hashes them in a compatible way). For instance, LLDAP doesn't expose the password hashes because it's doing a zero-knowledge proof that basically can't be replicated in gonic.

I totally agree on the security thing, it was a concern I had with the idea too, but my thought was that you could just hash the password after the request (considering gonic would already know it).  I still don't think it's a good idea though.  


> How does normal auth work in gonic? The credentials are sent with every request? 

Yeah, every request includes [URL parameters](https://opensubsonic.netlify.app/docs/api-reference/#authentication).   


> You can implement a server-side auth caching method, i.e. "these credentials were authorized by the LDAP server at time T, so they'll be valid until T+24h before we recheck them"

I was thinking about this, and the more I do the more logical it seems.  I don't know how @sentriz would feel, but the password reset button (in the web UI) could be replaced with an "LDAP refresh" button for LDAP users/in LDAP mode to clear the cache for that user.

betapictoris:
@sentriz, on implementing a cache for LDAP credentials, currently local user passwords are stored in plain text - which I think we can all agree is not a safe long-term solution.  Although, this raises the question as to how securing the cache should be approached, I can think of a few ways of doing this (but wanted to run them by you before doing anything that might not fit in with the project) including implementing:

- a more complex (and therefore secure) method of password storage (such as Argon2) for LDAP credentials, and share code between the database and cache down the line.
- a simple hashing method (such as MD5 or plain sha256) to go one or two steps beyond a plain text password, but still have a fairly simple and temporary solution.
- no hashing right now, and add it in once hashing is (maybe?) implemented for the database.

Also, if you don't want to add hashing yourself to the database, I might be able to add it later (although I don't have the resources to work on opening another pull request at the moment).

sentriz:
to my knowledge, storing is plaintext is a requirement from the subsonic protocol. this is to check credentials using the salt+token method

sentriz:
there is some more discussion here https://github.com/opensubsonic/open-subsonic-api/discussions/25

betapictoris:
> to my knowledge, storing is plaintext is a requirement from the subsonic protocol. this is to check credentials using the salt+token method

Sorry for the late response, but I hadn't thought of that, thanks for letting me know.

server/ctrlsubsonic/ctrl.go
betapictoris: Fixed in 6eddc2a1a9b2f1285cba4f757078e7f83627e818.  Thanks for the review!

server/ctrladmin/handlers.go
```diff
@@ -138,6 +138,125 @@ func (c *Controller) ServeUnlinkListenBrainzDo(r *http.Request) *Response {
 	return &Response{redirect: "/admin/home"}
 }
 
+func (c *Controller) ServeLDAPConfig(_ *http.Request) *Response {
```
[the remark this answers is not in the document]
betapictoris: This has been added in b36f26e32ad678d01114c04c538da9a850761f89.

server/ctrladmin/handlers_raw.go
betapictoris: Fixed in 6ca68b2c260d855b8965b762621efd9396c16107.

server/ctrlsubsonic/ctrl.go
betapictoris: Fixed in 6ca68b2c260d855b8965b762621efd9396c16107, and I've added an admin filter in 5638ef5df7346adf28335c0c125100004416e876.

betapictoris:
> You can implement a server-side auth caching method.

@nitnelave, this has been done in fccf01642f0553fd6d22336e14c8bb06d591c478.

server/ctrladmin/handlers_raw.go
betapictoris: Finished in f5195f15c4042357417d866af66ce8fd9fd711ee.

cmd/gonic/gonic.go
```diff
@@ -106,7 +125,7 @@ func main() {
 		log.Fatalf("invalid exclude pattern: %v\n", err)
 	}
 
-	if len(confMusicPaths) == 0 {
+	if len(confMusicPaths) == -1 {
```
sentriz, line 128: is this left over from testing or smth?

sentriz (commented):
hi there  seems like there some lint errors

cmd/gonic/gonic.go
betapictoris: Yeah, it was.  It's fixed in 7159e1ca09ba389674a3f23c2319d99906da58f5.

Issue #349, opened 2023-08-13
Convert podcast description to text


brian-doherty:
The Subsonic players I've used for podcasts (DSub, Airsonic) don't handle HTML in the podcast description well and so it doesn't render properly. Purpose instead to convert to text.

sentriz:
i think something like bluemonday could be a good fit for that 
![image](https://github.com/sentriz/gonic/assets/6832539/390d28fe-f056-4e1d-8e19-3050b22985cd)

brian-doherty:
Don't get me wrong -- it would be way cooler if the client would either
strip or (preferably) render the HTML. But other clients don't parse it
either.

Does the parser you suggest add in \n\n for the paragraph breaks? And
should it?

archekb:
>  it would be way cooler if the client would either strip or (preferably) render the HTML

brian-doherty:
Yeah, the issue is that other clients don't render the HTML either. I'm asking whether we should just strip at the server because of that. For example, DSub.

archekb:
I suppose it's not server issue (because server get description from podcast and sent it to client as raw), this is client issue, maybe you need to open issue for client?

Because after your comment I'll add support html tags to Airsonic-refix. ;)

brian-doherty:
So I checked and while the Subsonic API definition is ambiguous on this, the actual Subsonic implementation converts HTML to text on the way into the database. The magic question is -- do we keep the HTML and give clients the opportunity to go above and beyond, or do we match so that clients used to the way that Subsonic does things render correctly?

sentriz:
i think keep it simple for now - if we sanitise the html on the way out of the db and into the rest view, we keep all clients happy, while still keeping the data intact any clients decide to support it later

brian-doherty:
![Screenshot_20230814_143323_DSub](https://github.com/sentriz/gonic/assets/76168809/2640ced8-28b1-4383-b0ee-ea82c56a6b15)

So, bluemonday does OK but doesn't add carriage returns for paragraph breaks. See attached photo. I wonder if https://github.com/k3a/html2text or https://github.com/jaytaylor/html2text would be better.

brian-doherty:
![Screenshot_20230814_144957_DSub](https://github.com/sentriz/gonic/assets/76168809/c9b4d8fe-6091-4508-960e-34c6f8240fa3)
I used jaytaylor/html2text because it was more popular and I think it's way better. I'll send that in as a PR for you to see what you think.

[...]

PR #351, opened 2023-08-14
Added code to parse podcast episode descriptions from HTML to plain t…

brian-doherty:
…ext.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/securecookie v1.1.1
 	github.com/gorilla/sessions v1.2.1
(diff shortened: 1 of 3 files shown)


brian-doherty:
Fxies issue #349

sentriz:
thanks! did you have a look at bluemonday though? this html2text seems to implement its own html parser which maybe isn't the best idea? or maybe we don't really care in this case

also it outputs markdown. do we want that?

brian-doherty:
I tried bluemonday but it doesn't even leave a space between paragraph breaks, let alone a carriage return or two. This looks much better. FYI I chose the options to output text only rather than markdown.

brian-doherty:
See the photos I pasted in the issue for comparison with bluemonday.

PR #352, opened 2023-08-15
Add option for raw podcast description

archekb:
issue #349


brian-doherty:
Cool by me. I think "RAW" should be changed to "Raw" for stylistic reasons since it's not an acronym.

sentriz:
i'm a little bit hesitant here since we don't know of any subsonic clients that support html, and a like to have as few config options as possible

i know there is https://github.com/tamland/airsonic-refix/pull/110 but we could wait until it's merged?

or see if @opensubsonic would support a protocol to negotiate description text formats?

tamland:
Also hesitant about this as it can completely break client layout when tags such as hX, div and other are allowed. Also there's the never ending security issues with sanitization (case in point https://github.com/tamland/airsonic-refix/pull/110 contains the `target="_blank"` vulnerability)

Maybe if the html was limited to only tags that won't cause layout issues, but I'm not sure what that would be.

archekb:
Come on... Security problems ... sounds scared, but I don't think so it can be problem and this is can be fixed by adding attribute `rel="noopener noreferrer"`. Only h5 and h6 allowed, I can remove it, it's not a problem... We can create very conservative default option, but allow to users change it in client settings later.

archekb:
@sentriz I make a fork of airsonic-refix, it's named xonic-ui (https://github.com/archekb/xonic-ui), this client support this feature.

sentriz:
hey thanks for the effort here but closing for now until it's a supported feature from @opensubsonic . (xonic-ui is the only client that supports this right)

 feel free to re open then

PS will check out the share feature soon 👍

PR #353, opened 2023-08-15
Fix return playlist

### LINKED ISSUE #345: Get Playlist - error, if track was removed from library
@sentriz, How do you think, are we should reject playlist fully, if we delete some track from library (and rescan the library)? 
I suppose - not, we should do not return error for playlist, we should make a `continue` (for hide this track) or return playlist with this track, for user manual deletion or delete this track automatically, but in case if mount share was not mounted (in any reason), we do not should remove this track from list, because it can returning, then third variant is not valid, first variant not good too, because we will get many garbage in playlist, then the good choose is second variant, return full playlist for manual deleting.

https://github.com/sentriz/gonic/blob/8f05f33ab8a1534433a252c085cbd5586084dd09/server/ctrlsubsonic/handlers_playlist.go#L209-L212

archekb:
Fix return playlist if file not found and do not add to playlist not downloaded podcast.

fixes #345

### DIFF
--- a/server/ctrlsubsonic/handlers_playlist.go
+++ b/server/ctrlsubsonic/handlers_playlist.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"log"
 	"net/http"
 	"sort"
 	"time"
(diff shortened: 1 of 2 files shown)


server/ctrlsubsonic/handlers_playlist.go
```diff
@@ -237,5 +240,8 @@ func playlistRender(c *Controller, params params.Params, playlistID string, play
 		trch.TranscodedSuffix = transcodeSuffix
 		resp.List = append(resp.List, trch)
 	}
+
+	resp.SongCount = len(resp.List)
```
sentriz, line 244: i think this is a bit less brittle than decrementing the variable

PR #354, opened 2023-08-16
Parse podcast description from HTML to text just as with episodes.

brian-doherty:
If you're not going to take the change to make this a config option, then you should take this PR instead to parse podcast descriptions just as with podcast episodes, which is in that PR and is a good idea regardless IMO.


sentriz:
thanks, yes make sense until we have a decision on #352

PR #390, opened 2023-10-12
fix(subsonic): songCount and albumCount in genre objects is required

Duncaen:
The `songCount` and `albumCount` attributes are required according the subsonic api xml file. Noticed this in the airsonic-refix client where the genres page would sort the genres wrong if the value is left out instead of 0.


sentriz:
thanks!

PR #394, opened 2023-10-19
refactor(scanner): make the watcher a little easier to reason about

sentriz:
hiya @brian-doherty , i was reading the scanner code again today, and since it had been a while it took me a little while to grasp it again. i'm wondering what you think about these changes that i've found to help a me a bit

- pass only an "absPath" arg to scanCallback and watchCallback instead of "absPath" and "dir" (musicDir). the callbacks can deduce the root dir themselves, and the two args have always been confusing, even for the original scanCallback
- delete the watch stuff from the main scanner struct. it seems we don't need watchMap given above^ and the others can be centralised to ExecuteWatch
- update some names so that its clear we're doing some event batching

thoughts? thanks!

### DIFF
--- a/cmd/gonic/gonic.go
+++ b/cmd/gonic/gonic.go
@@ -300,12 +300,13 @@ func main() {
 
 		defer logJob("scan watcher")()
 
+		done := make(chan struct{})
 		errgrp.Go(func() error {
 			<-ctx.Done()
-			scannr.CancelWatch()
+			done <- struct{}{}
 			return nil
 		})
-		return scannr.ExecuteWatch()
+		return scannr.ExecuteWatch(done)
 	})
 
 	errgrp.Go(func() error {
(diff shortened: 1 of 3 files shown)


brian-doherty:
This mostly makes sense. My concern about dropping the passing of the root dir regards handling of sym links -- you can't deduce the root dir if you've followed a sym link to somewhere else in the file system. I don't use sym links in my library but you've got handling for them so they need to work properly.

sentriz:
oh yeah you're right :+1:

sentriz:
actually @brian-doherty it seems to be fine, have a look at this

image

when we encounter a symlink, we evaluate it and replace the prefix of the originally scanned path before evaluating

so if we scan a 
/mnt/music/artist (sym to external fs)
we evaluate to 
/mnt/external/artist
and carry on as if there was never a symlink in the first place
/mnt/music/artist/album

so the reverse mapping of music dirs works fine, as far as i can tell / test

brian-doherty:
Makes sense, as long as you tested it and it works.

sentriz:
did some more testing of symlinks seems to be fine :+1: 

thanks for your input, and the watcher itself :+1: :+1:

PR #424, opened 2023-12-10
Fix listenbrainz track length submission and include submission client details

phw:
This fixes the track length submission for ListenBrainz. ListenBrainz defines the fields `duration` and `duration_ms` for this, see https://listenbrainz.readthedocs.io/en/latest/users/json.html . Without this information LB will not show a track length. It might also prematurely remove the "now playing" notification for longer tracks.

As I understand from the code the track length is in seconds precision, hence the `duration` field is being used. `duration_ms`  would be for milliseconds.

This PR also includes the fields `submission_client` as defined in the LB API docs linked above.

### DIFF
--- a/listenbrainz/listenbrainz.go
+++ b/listenbrainz/listenbrainz.go
@@ -10,6 +10,7 @@ import (
 	"net/http/httputil"
 	"time"
 
+	"go.senan.xyz/gonic"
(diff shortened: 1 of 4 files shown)


sentriz:
thanks!

PR #428, opened 2023-12-16
add webui-path option and mux handler

tbmale:


tbmale:
I'm in no way a golang programmer but this is schratcing my itch and I'm using it for my local instance, so I tought...

sentriz:
what's the usecase for this?

tbmale:
> what's the usecase for this?

There are several subsonic web frontends out there ([jamstash](https://github.com/tsquillario/Jamstash),[airsonic-refix](https://github.com/tamland/airsonic-refix),[subplayer](https://github.com/peguerosdc/subplayer), etc.)

By doing ```-webapi-path``` I can self-host the frontend from gonic.

sentriz:
i would recommended having your reverse proxy with static file server host those. gonic can also read the GONIC_PROXY_PATH option to be aware of any path prefixes 

hosting from gonic may also make it responsible for any security implications there. such as hosting two sites on the same domain, one of which isnt sanitising user input

PR #443, opened 2023-12-29
listenbrainz: submit release MBID

phw:
This will include the MusicBrainz release ID in the request when submitting listens to ListenBrainz, see https://listenbrainz.readthedocs.io/en/latest/users/json.html#submission-json


sentriz:
thanks!

PR #469, opened 2024-02-13
Bug fix -- getAvatar would falsely return "user not admin" when user requested own avatar

brian-doherty:
This was because user and reqUser were deeply equal but different objects. Changed to check their IDs instead.


sentriz:
oh nice, thanks!

Issue #259, opened 2022-11-20
Allow Aliases for root music folders


Cherryblue:
Hi,

The New version of Gonic allows for multiple music folders, which is great for my personal usecase : 
- Having friends around a self hosted dropbox alternative (seafile)
- Each of my friends having their own music folders (not encrypted, contrary to the other folders)
- Telling Gonic to take in these music folders

While this seems perfect, I do not decide on the specific names of these folders. They show like this : 
- /friendA/7dd304f9-707b-4b32-a577-99303d82a326_Music
- /friendB/76ca605b-0d2d-429e-a036-dcde1ed3ec79_Musics

This isn't so great to see on my favorite music client.

**Enhancement : Allowing Aliases/Custom Names for "root folders" or "music sources" would be great.**
Either in the environment variable, or using gonic web page.

sentriz:
cool! good idea

now you can specify alias with a "->" syntax

for example

`gonic -music-path "music->/my/music" -music-path "friend one->/friend/one" -music-path "friend two->/friend/two`

or with env vars

`GONIC_MUSIC_PATH=music->/my/music,friend one->/friend/one,friend two->/friend/two
`

the syntax is a bit funny but it should do the trick. spaces are also fine

`my nice music -> /path/to/music`

Cherryblue:
Works like a charm, thank you Sentriz ! :D

PR #471, opened 2024-02-16
Update README.md following issue #259

Cherryblue:
Welp I observed readme had never been updated with this feature. This needs to be mentionned for everyone.


sentriz:
thanks! though i decided to integrate it better with what was already there. i don't think the features deserves as much vertical space as you gave it. also to nit, gonic is stylised with a lowercase letter

PR #479, opened 2024-03-06
Misc. small changes

ThinkChaos:
Hi,

I looked at the code and did some minor changes to get familiar.
Mostly lint/style fixes.

The main behavior change is listening on `127.0.0.1` only by default for added security, and since that aligns better with what packagers want.


sentriz:
hi , would you be able to make the 127.0.0.1 change separately please?

ThinkChaos:
Hey, I force pushed to the previous commit.

PR #487, opened 2024-03-23
feat: MPD server

ThinkChaos:
Hi,

I did this as a proof-of-concept. There's no docs, not enough tests and comments, and what's actually implemented from the MPD API is pretty minimal. So I've marked this as a draft and wanted to get feedback before spending any more time on it.

The end goal for me is to be able to control the jukebox via any MPD client, since the jukebox support of Subsonic clients, is subpar from my experience. Ideally in the end being able to steal the queue from device to device, like is possible on the streaming giants.

Please let me know if this is something you'd be interested in merging. I'm willing to handle maintenance/future issues that pop up.

### DIFF
--- a/server/mpd/server.go
+++ b/server/mpd/server.go
@@ -69,109 +69,160 @@ func (s *Server) Serve(ctx context.Context, l *net.TCPListener) error {
 	}
 }
 
-type connection struct { // TODO?: remove this type
-	*Server
-	*net.TCPConn
+type client struct {
+	srv *Server
+
+	lines *lineReader
+	out   io.Writer
+
+	cmdIdx  uint
+	cmdName string
 }
 
-func (c *connection) writeLine(line string) error {
-	n, err := io.WriteString(c.TCPConn, line+"\n")
-	_ = n
+func newClient(srv *Server, conn *net.TCPConn) *client {
+	lines := newLineReader(bufio.NewReader(conn))
+
+	return &client{
+		srv: srv,
+
+		lines: lines,
+		out:   conn,
+	}
+}
+
+func (c *client) writeLine(line string) error {
+	_, err := io.WriteString(c.out, line+"\n")
 	return err
 }
 
-func (c *connection) writePair(name, value string) error {
+func (c *client) writePair(name, value string) error {
 	return c.writeLine(fmt.Sprintf("%s: %s", name, value))
 }
 
-func (c *connection) handle(ctx context.Context) error {
-	defer c.Close()
-
-	c.writeLine(protocolHello)
-
-	lines := newLineReader(bufio.NewReader(c.TCPConn))
-
-	err := c.handleCmd(lines)
-	if err != nil {
-		var errRsp *errorResponse
-		if errors.As(err, &errRsp) {
-			c.writeLine(errRsp.String())
-			return nil
-		}
-
-		return err
-	}
-
-	return c.writeLine("OK")
+func (c *client) newErr(code ackError, err error) error {
+	return newError(code, c.cmdIdx, c.cmdName, err)
 }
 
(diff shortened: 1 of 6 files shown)


server/mpd/server.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+func doPause(c *client, args *argParser) error {
+	state, ok := args.Next()
+	switch {
+	case !ok: // no arg, toggle
+		return c.srv.jukebox.TogglePlay()
+
+	case state == "1" || state == "0":
+		return c.srv.jukebox.SetPlay(state == "0")
```
ThinkChaos, line 292: Rename `state` to `pause` to make the comparison clear.

server/mpd/server.go
```diff
@@ -0,0 +1,326 @@
...
+
+	default:
+		return c.newErr(ackErrorArg, fmt.Errorf("play state must be 0 or 1, got: %s", state))
```
ThinkChaos, line 295:
```suggestion
		return c.newErr(ackErrorArg, fmt.Errorf("pause state must be 0 or 1, got: %s", state))
```

server/mpd/server.go
```diff
@@ -0,0 +1,326 @@
...
+		"command_list_ok_begin": doCmdListOkBegin,
+		"currentsong":           doCurrentSong,
+		"pause":                 doPause,
+		"play":                  doPlay,
+		"status":                doStatus,
+	}
+}
+
+func doCmd(c *client, name string, args *argParser) error {
+	fmt.Println("->", args.line)
```
ThinkChaos, line 175: Needs to be removed/made into a proper log.

server/mpd/server.go
```diff
@@ -0,0 +1,326 @@
...
+	"io"
+	"log"
+	"net"
+	"net/netip"
+
+	"go.senan.xyz/gonic/jukebox"
+)
+
+const (
+	protocolVersion = "0"
```
ThinkChaos, line 17: Put something here when enough of the protocol is supported.

ThinkChaos (commented):
Just some comments to not forget things, since I took a quick look with fresh eyes.

lomereiter:
Why make it part of gonic though? It makes more sense to have it as a separate executable, that you can slap on top of any Subsonic-compatible server. This Mopidy plugin is looking for a maintainer, by the way: https://github.com/Prior99/mopidy-subidy

ThinkChaos:
@lomereiter Having a separate server seems a lot harder, and less optimized, since you can only use the public APIs, not reach for the DB/functions directly. And on top of that you now need to support/test multiple servers, not "just" multiple clients, so MxN vs N.

I am not interested in contributing to Mopidy because I don't want to use it (tried before Gonic FWIW), and because of tech and license reasons.

P.S. please consider how you phrase feedback because it sounds like you're dismissing the thought I put into this by using absolute wording like "it makes more sense", and telling me what to do with my time. Hopefully you can understand :)

lomereiter:
Fair enough, I just stated my opinion: I prefer the Unix way when every tool does one thing well. Sorry it came off as abrasive. And I'm not the maintainer, so it doesn't have much weight anyway.

I didn't say anything about having to support multiple servers, it could be just Gonic, and whoever needs support for other servers is free to contribute - that's how open source is supposed to work. Using public API is overhead indeed, but the benefit is that it's much more stable and should require almost no maintenance. Whereas if you want to consume Gonic as a library, it first needs a refactoring so that more or less stable interfaces are defined for accessing the DB instead of making queries directly from HTTP handlers, as is the case now (see server/ctrlsubsonic).

maristieuvelonde:
This is a fantastic idea, I would definitely use this!

Issue #306, opened 2023-03-15
REQ: Server-side playlist import?


xxxserxxx:
It's common among servers to have a directory in which `m3u` playlists are stored as files; this location is specified by a path. LMS stores user-created playlists (with absolute filesystem paths) in this directory, and loads them from there. Navidrome and mpd will also pick up playlists from this directory.

If gonic supports this, I've missed it. I didn't see an environment variable that would point gonic to the playlist directory, nor did I find code that would load `m3u` playlists from a directory in a quick scan. I may have missed it, though.

If gonic can't do this thing, I'd like to request it. gonic already has `m3u8` playlist support; to support a playlists directory, gonic would need:

- A configuration variable for the user to provide the path to the playlists, e.g. `GONIC_PLAYLIST_PATH`
- Code to read in files from that directory
- Code to trigger a rescan of the directory (I think most simple would be hooking into the existing "scan for changes" template function)

sentriz:
interesting idea

maybe it could even be taken a step further to make things simpler, like a directory `/playlists/user/<playlist name>.m3u8`

then when a getPlaylist or getPlaylists etc called, we just read that directory and generate a response on the fly. it would be plenty fast

then there would be no need for importing, scanning, or any syncing with the database. and we would also solve #54 for free

thoughts @brian-doherty ?

brian-doherty:
Well playlists need IDs so any way you look at it I think you need to involve the DB. I think xxxserxxx's proposal makes sense given that those imported playlists become public playlists available to all users.

Oddly, the Subsonic API spec has a notion of a list of allowed users for playlists (http://www.subsonic.org/pages/inc/api/examples/playlists_example_1.xml) but no API to set that list of users. Under the API playlists are either personal to the user or public for all.

brian-doherty:
You could. It would make the playlist IDs out of sync with the rest of the ID system. You'd also have to convert your existing playlist IDs to the new system. I suspect it'd be less work to use the existing logic to just import them.

xxxserxxx:
I have a patch that implements this; I'm testing it now. Basically, it:

- Adds a GONIC_PLAYLIST_PATH
- Reads all of the files ending in `.m3u` or `.m3u8`, and for each one of those,
- calls `ctrladmin.playlistParseLine()`

To enable this, I had to move some code from `ctrladmin` to `scanner`, but I think it makes sense. You could even call it solving some feature envy on the part of `ctrladmin`... but you can make your own call on that when you see the patch.

[...]

PR #499, opened 2024-04-24
Add playlists with no user to admin's playlist list.

xxxserxxx:
This is carrying on from the various tickets under which playlists have been discussed.

For reasons well-known (#306, #311, #307, #311... but mainly #306), the Subsonic API forces playlists to be by user ID. The solution in gonic was to expect a directory layout with `<playlistpath>/<userid>/.../<whatever>.m3u`. This is inconvenient for users for whom gonic is not the only application using the playlist directory, and maybe not even the main one, an inconvenience exacerbated by the fact that `fs.DirEntry.IsDir()` returns false for directory symlinks.

This extremely trivial change does one thing: instead of throwing an error when the directory structure does not contain user IDs as the second element in the path, it merely adds those playlists to the admin's playlist list. This change has two impacts:

1. These playlists would otherwise be entirely ignored by gonic. It's possible some users are leveraging that to have playlists in the same directory used by some other program that they intentionally hide from gonic.
2. admin might suddenly see a bunch of playlists they didn't before, because there were playlists that were being ignored

What it accomplishes is that it allows gonic to share playlists with all of the other non-Subsonic apps that _don't_ put playlists into int-named subdirectories.

### DIFF
--- a/playlist/playlist.go
+++ b/playlist/playlist.go
@@ -99,7 +99,7 @@ func (s *Store) Read(relPath string) (*Playlist, error) {
 
 	playlist.UserID, err = userIDFromPath(relPath)
 	if err != nil {
-		return nil, fmt.Errorf("convert id to str: %w", err)
+		playlist.UserID = 1
 	}
 
 	playlist.Name = strings.TrimSuffix(filepath.Base(relPath), filepath.Ext(relPath))


sentriz:
interesting. what happens then when the admin goes to edit the playlist then

sentriz:
havent tried but I think it end up making a copy of it or something when you go to update

xxxserxxx:
I don't know; I'm not using a client that allows editing the playlist, but from a quick scan of the code CreateOrSave handler reloads all the playlists and only constructs a new file path if it doesn't find a matching playlist. I believe if Admin edits an existing playlist from the parent playlist directory, it'll actually edit the playlist and not create a new one.

c.f. `server/ctrlsubsonic/handlers_playlist.go#78`, where it tries to load the playlist by ID using this same `Load()` function. The user ID doesn't come into play unless the playlist is new, in which case `SaveCreateOrUpdatePlaylist()` creates a new playlist path using the user ID on line 104.

What I'm not sure about is what happens when Gonic encounters duplicate playlists names -- although, this is already an issue if 3 different users all create playlists with the same name, and mark them all GLOBAL. I think the real issue is that if users are creating playlists in tools other than a Subsonic client, playlists will not be assigned IDs, but they'll

PR #512, opened 2024-06-08
fix sort for dir, tracks

mikuta0407:
In the current implementation, if the file names are as follows, they will be output in the expected order when the client displays the list.

```
01 aaa.mp3
02 bbb.mp3
...
10 jjj.mp3
11 kkk.mp3
...
20 ttt.mp3
21 uuu.mp3
```

However, if the leading zero is not as shown below, iSub and Jamstash will sort by string instead of by track number.

```
1 aaa.mp3
2 bbb.mp3
...
10 jjj.mp3
11 kkk.mp3
...
20 ttt.mp3
21 uuu.mp3
```
↓
```
1 aaa.mp3
10 jjj.mp3
11 kkk.mp3.
...
2 bbb.mp3
20 ttt.mp3
21 uuu.mp3
```

This was resolved by inserting sort.

Also, DSub (Android) and others have a function to sort and display albums based on the release year value in the album list, but since `Year` is not stored when NewTCAlbumByFolder is executed in the `GET /getMusicDirectory` process, the value is not included in the HTTP response and year sorting is not possible, so this has been fixed.


mikuta0407:
Oops... the test failed... I'll look into fixing that later.

sentriz:
maybe we could just add an extra Order() to sort?

mikuta0407:
As you pointed out, adding the order solved the problem. I added a commit in the fixup, so if there are no problems, I will squash it.
https://github.com/mikuta0407/gonic/commit/e0931132d92ef39a8d9e60e618e7a062a8a4ed1e

PR #516, opened 2024-06-11
feat: improve cover selection algorithm

heimoshuiyu:
Hi,

Gonic is a fantastic project for local music streaming. However, I noticed that Gonic doesn't recognize some album covers. After digging into the code, I found that cover art can only be recognized if it follows certain fixed patterns. Most of my albums are downloaded from BitTorrent, so the naming of cover art varies. Here is a sample file list created by running `find /data/ | grep -i -E '.png|.jpg|.jpeg|.webp|.gif '`:

[files.txt](https://github.com/user-attachments/files/15786788/files.txt)

To address this issue, I propose a simple but more effective way to select cover art from an album. Please check the code for details.

A docker images built on this PR is available: `docker.io/heimoshuiyu/gonic:cover`

### DIFF
--- a/scanner/coverresolve/cover.go
+++ b/scanner/coverresolve/cover.go
@@ -1,0 +1,83 @@
+package coverresolve
+
+import (
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+var DefaultKeywords = []string{
+	"cover",
+	"folder",
+	"front",
+	"albumart",
+	"album",
+	"artist",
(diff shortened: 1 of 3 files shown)


sentriz:
thanks! i needed this for another project (this project will provide some lower level audio metadata and related functionality for Go) I implemented it there, and used your test cases. hoep you dont mind. thanks!

https://github.com/sentriz/wrtag/blob/master/coverparse/coverparse.go

PR #488, opened 2024-03-24
feat: implement lyrics

TorchedSammy:
closes #371 

Implements both /getLyrics and OpenSubsonic [/getLyricsBySongId](https://opensubsonic.netlify.app/docs/endpoints/getlyricsbysongid/)

TODO:
- [ ] /getLyrics
  - [ ] Use artist in search for songs
  - [ ] Embedded
  - [x] .lrc
- [ ] /getLyricsBySongId
  - [ ] Embedded
  - [x] .lrc

PR #519, opened 2024-06-19
Add support for lyrics

danielepintore:
In this pull request I've added support for lyrics.

When we make a scan now we check for lyrics inside the file using `taglib`, once the scan process in finished we have all lyrics saved on the db so we can retrieve them faster. When we are looking for a lyrics it first checks if there is a embedded lyric, if there is we return that, otherwise it looks for a file `SONG_NAME.lrc` and reads the lyrics from it. If there isn't a `.lrc` file an empty response is provided. 

I've added support for both `getLyrics` and `getLyricsBySongId` methods. I have even created a migration, but i'm not so sure if that works fine maybe give that a double check.

Let me know if there is something to change.
P.S. I have used some code from #488


sentriz:
thanks for work here, but is this a dupe of https://github.com/sentriz/gonic/pull/488 ?

danielepintore:
Not exactly a dupe since I've added also the possibility of fetching the lyrics from the embedded tag. Since i couldn't work on that pull request I've decided to make a new one.

gfififid:
How can i test it? I can't understand how to build it

danielepintore:
Hi, I suggest to build a custom docker image and use that in your `docker-compose.yaml`.

1. Clone the repo: `git clone git@github.com:danielepintore/gonic.git`
2. `cd gonic`
3. Build the image: `docker build . -t gonic_lyrics:latest`
4. Replace in the docker-compose the image name with `gonic_lyrics:latest`
5. Run the container the same way you do for the standard version `docker compose up -d`

These are the steps needed to build and use the image, but i have noticed that it returns an error. The problem is the one fixed by #526. Changing the Dockerfile with the changes proposed by #526 fixes the issue and let you build the image.

gfififid:
That's works. Thank you!

BANanaD3V:
Any updates?

danielepintore:
The feature works fine, I've been using it for a few months without any problems. We should wait for a response from @sentriz

xxxserxxx:
For client testing, I just pushed a branch onto [@spezifisch 's fork of stmps](https://github.com/spezifisch/stmps/tree/lyrics) that implements showing synchronized lyrics fetched from @danielepintore 's fork.

It's functional. stmps uses mpv as the player, and gets time sync's every second, which means lyrics timings are always just a little off. I found it's better to switch lyric lines early, rather than later, so that the lyric line is shown when the vocals happen; anyway, it works.

I'd be thrilled to see @danielepintore 's fork merged; now there's one more client that can use it!

server/ctrlsubsonic/handlers_common.go
```diff
@@ -468,12 +471,154 @@ func (c *Controller) ServeJukebox(r *http.Request) *spec.Response { // nolint:go
...
+		times, lrc, err := lrc.ParseString(track.Lyrics)
+		if err != nil {
+			return spec.NewError(0, fmt.Sprintf("lyricsFile: %v", err.Error()))
+		}
+
+		lines := make([]spec.Lyric, len(times))
+		for i, time := range times {
+			lines[i] = spec.Lyric{
+				Start: time.Milliseconds(),
+				Value: lrc[i],
```
DatGizmo, line 557: Not sure if it is standard or not, but the lrc source files I have have a space between the time tag and a string.
`[mm:ss] Line`

I'm using Tempo as my player on Android and there the lyrics highliter would shift by a letter with each new line.

I checkd with the api and noticed that the `value` would start with a leading space
`"value": " Line"`

I added a `string.TrimSpace` around the `lrc[i]` and that fixed the space offset issue with Tempo

server/ctrlsubsonic/handlers_common.go
```diff
@@ -468,12 +471,154 @@ func (c *Controller) ServeJukebox(r *http.Request) *spec.Response { // nolint:go
...
+			return sub
+		}
+		return spec.NewError(0, fmt.Sprintf("lyricsFile: %v", err.Error()))
+	}
+
+	lines := make([]spec.Lyric, len(times))
+	for i, time := range times {
+		lines[i] = spec.Lyric{
+			Start: time.Milliseconds(),
+			Value: lrc[i],
```
DatGizmo, line 591: Same here :)

server/ctrlsubsonic/handlers_common.go
danielepintore: If i remember correctly there shouldn't be any space, but i think that we should trim them anyway

danielepintore:
Hey @sentriz any update on this getting merged? I've just rebased to master and I think it's ready to be merged

xxxserxxx:
FWI I have a hard fork of stmps that has support for lyrics (and cover art, and song metadata) that depends on (the lyrics part) this feature; my fork is in AUR and Alpine testing. I'm waiting on this being merged before moving the package from Alpine testing to community -- users can use Navidrome, which has this support, but since I use gonic I'm gating on this.

sentriz:
thank you @danielepintore and sorry for the delay!

Issue #501, opened 2024-05-12
Config setting to limit cache size


brian-doherty:
This is a feature request. Now that there is a transcoder cache, the cache can in fact take up quite a lot of disk space, but there's no mechanism to eject from the cache at any point. I recommend the addition of a config setting that can limit the size of the cache to a given amount of disk space, with LRU ejection using last access time.

PR #520, opened 2024-06-22
Added config option to set size of transcode cache and cadence to enf…

brian-doherty:
…orce that sizing via ejection.

### DIFF
--- a/transcode/transcoder_caching.go
+++ b/transcode/transcoder_caching.go
@@ -20,6 +20,7 @@ type CachingTranscoder struct {
 	transcoder Transcoder
 	limitMB    int
 	locks      keyedMutex
+	cleanLock  sync.RWMutex
 }
 
 var _ Transcoder = (*CachingTranscoder)(nil)
@@ -29,6 +30,9 @@ func NewCachingTranscoder(t Transcoder, cachePath string, limitMB int) *CachingT
 }
 
 func (t *CachingTranscoder) Transcode(ctx context.Context, profile Profile, in string, out io.Writer) error {
+	t.cleanLock.RLock()
+	defer t.cleanLock.RUnlock()
+
 	// don't try cache partial transcodes
 	if profile.Seek() > 0 {
 		return t.transcoder.Transcode(ctx, profile, in, out)
@@ -70,6 +74,9 @@ func (t *CachingTranscoder) Transcode(ctx context.Context, profile Profile, in s
 }
 
 func (t *CachingTranscoder) CacheEject() error {
+	t.cleanLock.Lock()
+	defer t.cleanLock.Unlock()
+
 	// Delete LRU cache files that exceed size limit. Use last modified time.
 	type file struct {
 		path string
@@ -106,9 +113,7 @@ func (t *CachingTranscoder) CacheEject() error {
 		curFile := files[0]
 		files = files[1:]
 		total -= curFile.info.Size()
-		unlock := t.locks.Lock(curFile.path)
 		err = os.Remove(curFile.path)
-		unlock()
 		if err != nil {
 			return fmt.Errorf("remove cache file: %w", err)
 		}
--- a/transcode/transcoder_caching.go
+++ b/transcode/transcoder_caching.go
@@ -7,7 +7,9 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sort"
 	"sync"
+	"time"
 )
 
 const perm = 0o644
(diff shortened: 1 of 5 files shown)


brian-doherty:
Resolves #501

brian-doherty:
@sentriz Would love to get your feedback on this when you get a minute. Thanks.

sentriz:
hiya Brian sorry for the delay . will hopefully take a look shortly

zpuskas:
@brian-doherty Could you please also update contrib/config accordingly with this new option?

brian-doherty:
> @brian-doherty Could you please also update [contrib/config](https://github.com/sentriz/gonic/blob/master/contrib/config) accordingly with this new option?

Done. Lint and test are running now but will pass.

brian-doherty:
@sentriz I know you're busy but I would love you to have a look at this PR when you get a sec.

transcode/transcoder_caching.go
```diff
@@ -64,6 +68,41 @@ func (t *CachingTranscoder) Transcode(ctx context.Context, profile Profile, in s
...
+
+	sort.Slice(files, func(i, j int) bool {
+		return files[i].info.ModTime().Before(files[j].info.ModTime())
+	})
+
+	for total > int64(t.limitMb)*1024*1024 {
+		curFile := files[0]
+		files = files[1:]
+		total -= curFile.info.Size()
+		unlock := t.locks.Lock(curFile.path)
```
sentriz, line 100: the t.locks is a keyedMutex but the keys are profile strings (ffmpeg commands with args and paths) not paths. so this has little effect i think

maybe we should have a global clean lock which the normal transcode thing respects too?

something like

```
type CachingTranscoder struct {
	...
	locks      keyedMutex
    cleanLock  sync.RWMutex
}

...

func (t *CachingTranscoder) Transcode(...) error {
    ...

    t.cleanLock.RLock()
    defer t.cleanLock.RUnlock()
    
    ...

	key := cacheKey(name, args)
	unlock := t.locks.Lock(key)
	defer unlock()
}

func (t *CachingTranscoder) CacheEject() error {
    t.cleanLock.Lock()
    defer t.cleanLock.Unlock()
}
```

that way the cleaning process locks the entire cache, while multiple goroutines looking to read the cache work fine, just not while we're cleaning

wdyt?

sentriz (commented):
looks good! just a couple things

transcode/transcoder_caching.go
```diff
@@ -7,21 +7,24 @@ import (
...
 	transcoder Transcoder
+	limitMb    int
 	locks      keyedMutex
 }
 
 var _ Transcoder = (*CachingTranscoder)(nil)
 
-func NewCachingTranscoder(t Transcoder, cachePath string) *CachingTranscoder {
-	return &CachingTranscoder{transcoder: t, cachePath: cachePath}
+func NewCachingTranscoder(t Transcoder, cachePath string, limitMb int) *CachingTranscoder {
```
sentriz, line 26: looks megabits to me. how about a big B for bytes

```suggestion
func NewCachingTranscoder(t Transcoder, cachePath string, limitMB int) *CachingTranscoder {
```

transcode/transcoder_caching.go
```diff
@@ -64,6 +68,41 @@ func (t *CachingTranscoder) Transcode(ctx context.Context, profile Profile, in s
 	return nil
 }
 
+func (t *CachingTranscoder) CacheEject() {
```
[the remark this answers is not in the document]
brian-doherty: Done.

transcode/transcoder_caching.go
```diff
@@ -64,6 +68,41 @@ func (t *CachingTranscoder) Transcode(ctx context.Context, profile Profile, in s
...
+	// Delete LRU cache files that exceed size limit. Use last modified time.
+	type file struct {
+		path string
+		info os.FileInfo
+	}
+
+	var files []file
+	var total int64 = 0
+
+	_ = filepath.Walk(t.cachePath, func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
brian-doherty: Done.

transcode/transcoder_caching.go
brian-doherty: Good call and my apologies. Done.

transcode/transcoder_caching.go
brian-doherty: Done.

brian-doherty:
@sentriz All comments addressed. See what you think and thanks!

cmd/gonic/gonic
sentriz, line 1: looks like this was committed accidentally

brian-doherty: My bad. Fixed.

sentriz:
looks great, thanks! sorry for the delay

Issue #522, opened 2024-06-26
Playlist sharing across users


roobre:
gonic version: v0.16.4

Hi there!
Long time lurker of the Gonic project here, which has now regained some of my interest as recent changes to Jellyfin have unfortunately impacted a core use case for me: playlist sharing. I'm very much not a C# person myself, but I can handle Go just fine. Because of this, I would like to pitch this feature and offer my help to get it done in Gonic!

My current idea is very minimal: Allow a (potentially small) set of users to control playlists they do not own.

I've taken a quick look at the Gonic codebase, and I think this could be implemented on a first iteration by defining a playlist attribute in the likes of

```
#GONIC-SHARED-WITH 1,3,5
```

I think this should play nice with the existing attribute parser: https://github.com/sentriz/gonic/blob/0e45f5e84cd650211351179edf3eed89a54c6c75/playlist/playlist.go#L246

With this list unmarshalled, it should hopefully be simple to tap into that list for consuming in 
https://github.com/sentriz/gonic/blob/0e45f5e84cd650211351179edf3eed89a54c6c75/server/ctrlsubsonic/handlers_playlist.go#L38

and for editing it in 
https://github.com/sentriz/gonic/blob/0e45f5e84cd650211351179edf3eed89a54c6c75/server/ctrlsubsonic/handlers_playlist.go#L133

To keep the implementation small I was thinking about not exposing a method for granting access to playlists on the first iteration, and instead instruct users to add this comment manually to their m3u files.

Would love to hear your thoughts on this! If you think this could be useful, I'll be happy to spin up a PR with the suggested approach.

jesperp:
The way I do this is simply add a symbolic from the users that wants access to a certain playlist. After all, they are just plain old files. Less code (no code) and it just works 😸

PR #524, opened 2024-07-03
Fix non-admin users not being able to create playlists

roobre:
Found this while working on a tentative implementation for #522. It seems that https://github.com/sentriz/gonic/pull/499 introduced some logic that prepopulates `playlist.UserID` to `1` if nothing else can be decoded from the path.

Before this PR, `ServeCreateOrUpdatePlaylist` was trying to figure out if a playlist existed by trying to open the file derived from its name. If the name does not exist, `playlistIDDecode` returns an empty string, which `playlistStore.Read` gladly accepts and proceeds to open `""`, which implies `.`, which does not fail. Then, the code added in #499 will slap the admin user ID on the playlist object, shortly before `playlistStore.Read` realizes this is actually a directory and returns an error.

That error, however, was never checked. As a result, an empty non-existing playlist with `UserID: 1` was returned, and prevented creating playlists as every non-existing playlist will "belong" to admin, and thus no one other than admin could "modify" it.

This PR adds a couple safety checks so this does not happen:
- Does not bother opening a playlist if the decoded path is empty
- Checks error returned by `Open`
- Causes `Open` to fail earlier if a path to a directory is passed

### DIFF
--- a/playlist/playlist.go
+++ b/playlist/playlist.go
@@ -94,6 +94,10 @@ func (s *Store) Read(relPath string) (*Playlist, error) {
 		return nil, fmt.Errorf("stat m3u: %w", err)
 	}
 
+	if stat.IsDir() {
+		return nil, errors.New("path is a directory")
+	}
+
 	var playlist Playlist
 	playlist.UpdatedAt = stat.ModTime()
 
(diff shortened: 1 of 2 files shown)


sentriz:
thanks!

PR #525, opened 2024-07-03
Allow sharing playlists across multiple users

nadiamoe:
This is my take at #522, implementing the easiest functional approach to sharing playlists by specifying a list of users that can edit another user's playlist.

For now, there is no UI for "inviting" users, the m3u file needs to be edited by hand. Users with whom a playlist has been shared can add/remove items from it, but not delete the playlist itself.

Includes commits from #524

### DIFF
--- a/playlist/playlist.go
+++ b/playlist/playlist.go
@@ -7,6 +7,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


playlist/playlist.go
```diff
@@ -122,6 +140,21 @@ func (s *Store) Read(relPath string) (*Playlist, error) {
 			playlist.Comment = value
 		case attrIsPublic:
 			playlist.IsPublic, _ = strconv.ParseBool(value)
+		case attrSharedWith:
+			sharedWith := strings.Split(value, ",")
```
roobre, line 144: Currently thinking whether it would be better for this to be a comma-separated list (as it is now), or if we should use the comment multiple times 🤔
```
#GONIC-NAME:"Roobre"
#GONIC-COMMENT:""
#GONIC-IS-PUBLIC:"false"
#GONIC-SHARED-WITH:"2"
#GONIC-SHARED-WITH:"3"
```
The latter sounds easier to parse.

PR #537, opened 2024-09-07
Add `-playlists-relative` option

### LINKED ISSUE #529: custom prefix for paths inside playlist
I like that gonic stores everything as files, but the playlist files it writes to use absolute paths which might change between machines.

It'd be nice if you could specify a custom prefix. For example, I have the following directory structure:

```
Music/
├── Playlists/
│   └── 1
│           └── playlist1.m3u8
├── Podcasts/
└── Artist 1
       └── Album 1
           ├── 01.song.flac
           └── folder.jpg
```

It would be nice if I could set the prefix to be `../../` in this case so that the playlists would work even if the absolute path of `Music/` changed.

0chroma:
Fixes  #529

Improves portability between machines by optionally making file paths relative to the playlist

feedback welcome! it's my first time writing go

### DIFF
--- a/cmd/gonic/gonic.go
+++ b/cmd/gonic/gonic.go
@@ -148,7 +148,7 @@ func main() {
 		DBPath:            *confDBPath,
 		OriginalMusicPath: confMusicPaths[0].path,
 		PlaylistsPath:     *confPlaylistsPath,
-		PlaylistsRelative: confPlaylistsRelative,
+		PlaylistsRelative: *confPlaylistsRelative,
 		PodcastsPath:      *confPodcastPath,
 	})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


sentriz:
hm I wonder do we need the config option at all? can we just allow relative paths? joining on the basedir if the playlist line doesn't look absolute with no leading slash

0chroma:
the option mostly determines if the playlist is written with absolute or relative paths, I modified the reader so it does exactly what you described

0chroma:
If you'd like I can remove the config check during writes from the PR, but I think there's a lot of situations where absolute paths would still make sense (ie having playlists and music on separate volumes)

0chroma:
heya, just checking in. Let me know if you'd like this merged in!

PR #541, opened 2024-09-25
Use file name / foder name for unknown tag retrieval.

heimoshuiyu:
When taglib cannot retrieve the title of the media or the title of the album, use the filename as the track name and the folder name as the album name.

I have a lot of music downloaded from YouTube, which doesn't have recognizable media metadata. After scanning with gonic, all of them turn into empty titles and unknown albums, making it difficult for me to distinguish the music. For example, like this:

Screenshot_20240925_150357

These pieces of music without metadata should be manually managed by the user using filenames and folders. In this way, using the filename and folder as the title and album name is reasonable. For example, like this: (the file is named `aino.opus` under folder `miku`)

Screenshot_20240925_150148

### DIFF
--- a/mockfs/mockfs.go
+++ b/mockfs/mockfs.go
@@ -371,6 +371,8 @@ func (i *TagInfo) ReplayGainAlbumPeak() float32 { return 0 }
 func (i *TagInfo) Length() int  { return firstInt(100, i.RawLength) }
 func (i *TagInfo) Bitrate() int { return firstInt(100, i.RawBitrate) }
 
+func (i *TagInfo) AbsPath() string { return "" }
+
 var _ tagcommon.Reader = (*tagReader)(nil)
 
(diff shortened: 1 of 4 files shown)


sentriz:
interesting, i think i'd need to weigh up the pros and cons here. this is what the subsonic "browse by folder" mode is for. have you tried that?

heimoshuiyu:
Yes, I know. However, the PC client `Sonixd` that I use does not support this mode. The `Ultrasonic` client on Android can choose this mode. The "browser by folder" effect is like this (without this PR):

Screenshot_2024-09-25-16-54-16-271_org moire ultrasonic

I don't think this PR has any drawbacks. It supplements the missing song information in ID3 mode, turning blank titles and Unknown Album into understandable information. At the same time, it does not change the behavior in "browser by folder" mode. I have been using it for two months and think this is a very good improvement.

PR #545, opened 2024-11-15
Initial implementation of a sandbox for Linux & OpenBSD

WickedShell:
Leverages pledge and unveil, and leaves a public API for other systems to follow. The API was designed to match the OpenBSD side pretty closely as that's the initial target, if a BPF/capsicum implementation is brought forward it may be worth changing the API, and we should be okay with that. (Neither of those targets are on my current list of things to actively implement).

I've been using this locally for a bit on an OpenBSD system, can confirmed that a Linux build still works fine on a quick test. I've invoked transcoding, the only major feature I haven't touched that I'm tracking is jukebox, which I'm not really sure how it's supposed to be working yet. I also was unsure how to invoke the database upgrade path that I can ensure works.

This is the first time I've done anything with Go, so any criticisms are welcome.


WickedShell:
I've now extended this to support landlock on Linux, and retested on OpenBSD.

@sentriz any opinions on this?

Issue #546, opened 2024-11-15
nil pointer panic when adding a podcast with no image


superfrink:
gonic version: v0.16.4

Adding the podcast `http://feeds.feedburner.com/thechangelog` causes a panic:
```
gonic[8916]: 2024/11/15 10:19:18 http: panic serving 100.1.5.56:50349: runtime error: invalid memory address or nil pointer dereference
gonic[8916]: goroutine 3514 [running]:
gonic[8916]: net/http.(*conn).serve.func1()
gonic[8916]:         /usr/local/go/src/net/http/server.go:1947 +0xbe
gonic[8916]: panic({0xc14360?, 0x12cc360?})
gonic[8916]:         /usr/local/go/src/runtime/panic.go:785 +0x132
gonic[8916]: go.senan.xyz/gonic/podcast.(*Podcasts).AddNewPodcast(0xc0002da360, {0xc000ae1860, 0x28}, 0xc00071c000)
gonic[8916]:         /root/go/pkg/mod/go.senan.xyz/gonic@v0.16.4/podcast/podcast.go:85 +0x17a
gonic[8916]: go.senan.xyz/gonic/server/ctrladmin.(*Controller).ServePodcastAddDo(0xc00028cf80, 0xc000582508?)
gonic[8916]:         /root/go/pkg/mod/go.senan.xyz/gonic@v0.16.4/server/ctrladmin/handlers.go:416 +0x228
gonic[8916]: go.senan.xyz/gonic/server/ctrladmin.respHandler.func2.1({0xe600a0, 0xc000b68690}, 0xc000241b80)
gonic[8916]:         /root/go/pkg/mod/go.senan.xyz/gonic@v0.16.4/server/ctrladmin/ctrl.go:210 +0x78
gonic[8916]: net/http.HandlerFunc.ServeHTTP(0xc000a285a0?, {0xe600a0?, 0xc000b68690?}, 0xc000047610?)
gonic[8916]:         /usr/local/go/src/net/http/server.go:2220 +0x29
gonic[8916]: go.senan.xyz/gonic/server/ctrladmin.withAdminSession.func1({0xe600a0, 0xc000b68690}, 0xc000241b80)
gonic[8916]:         /root/go/pkg/mod/go.senan.xyz/gonic@v0.16.4/server/ctrladmin/ctrl.go:177 +0x15f
[...]

PR #547, opened 2024-11-15
fix(podcast): nil pointer when podcast has no image

superfrink:
fixes https://github.com/sentriz/gonic/issues/546


sentriz:
thanks! added a commit to it

PR #554, opened 2024-12-26
fix(subsonic): allow searches without query

reennix:
Some subsonic clients rely on searches without a query to get all available content. This pull request ignores the `ErrNoValues` error to make empty searches work.


sentriz:
I think the semi standard thing to do there is search with "*"

sentriz:
that's what symfonium does

PR #567, opened 2025-02-24
fix(getPlaylist): populate ArtistID and AlbumID

richardsamuels:
The [OpenSubsonic getPlaylist](https://opensubsonic.netlify.app/docs/endpoints/getplaylist/) spec doc lists `artistId` and `albumId` as fields for playlist entries, which are currently not being populated. 

This PR populates those fields.

### DIFF
--- a/server/ctrlsubsonic/spec/construct_by_folder.go
+++ b/server/ctrlsubsonic/spec/construct_by_folder.go
@@ -131,6 +131,9 @@ func NewTCPodcastEpisode(pe *db.PodcastEpisode) *TrackChild {
 		IsDir:       false,
 		Type:        "podcastepisode",
 		CreatedAt:   pe.CreatedAt,
+		Album:       pe.Album,
+		Artist:      pe.Artist,
+		CoverID:     pe.SID(),
 	}
 	if pe.Podcast != nil {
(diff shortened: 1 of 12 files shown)


sentriz:
thanks @richardsamuels would you mind rebasing? i also usually format with `prettier --parser json --write server/ctrlsubsonic/testdata/*`

server/ctrlsubsonic/spec/construct_by_folder.go
```diff
@@ -101,6 +102,9 @@ func NewTCTrackByFolder(t *db.Track, parent *db.Album) *TrackChild {
 		trCh.Genres = append(trCh.Genres, &GenreRef{Name: g.Name})
 	}
 	for _, a := range t.Artists {
+		if a.Name == t.TagTrackArtist {
```
sentriz, line 105: what's this check for?

richardsamuels: Each `TrackChild` has a single `ArtistID`, which corresponds to the `TagTrackArtist`.  The check just makes sure that we assign the correct `ArtistID` to to the `TrackChild`. If this check wasn't here, it would assign the track's `ArtistID` to the last artist in the `Artists` list

Issue #565, opened 2025-02-17
Podcast missing data in response


bensmith2697:
gonic version: lastest

I hope I explain myself well here, but looking at the response of the `/rest/getNewestPodcasts` it doesn't return the podcast `title` (`album` in track), in this case it would be the name of the album. It only returns the `parent`. 

The same applies for when a podcast episode is added to the playlist, it doesn't return the same data as a track. Should it not return the same format?

Subsonic returns the `album`/`albumId` format but it doesn't in Gonic from what I can see.

bensmith2697:
@sentriz is this something that can be done?

bensmith2697:
@sentriz seems there is a pull request opened. Any chance you could take a look? I would like to see this fixed.

bensmith2697:
@sentriz can we re-open this it seems that this issue is not fixed, I am still getting empty values for artists. It seems [this feed](https://badfaith.libsyn.com/rss) and others I have only have an author at the main feed and not per episode.

![Image](https://github.com/user-attachments/assets/0d36b27c-7864-4236-8e6c-9e01c84e5db9)

sentriz:
hi @bensmith2697 did you run from source? the included fix has not been released yet

bensmith2697:
@sentriz is this not in the nightly release (I see I put latest when it should have been nightly)?

sentriz:
ah i see, can you try delete and readd the podcast

bensmith2697:
@sentriz I did this and it is still the same issue. When parsing the data if authors is not in the episode xml element, then it will be empty string, but it should use the podcast author as default if not set.

sentriz:
@VD39 any ideas?

VD39:
@sentriz @bensmith2697 I've noticed that this issue occurs with some of my podcasts as well. During my initial testing, I assumed it was mandatory for each episode, which is why I didn't test multiple XML feeds. It seems we will need to pass all the podcast details and verify them accordingly. I'll try and work on addressing this issue in the coming days.

bensmith2697:
@VD39 Thank you.

PR #569, opened 2025-03-03
Podcast episodes updates.

VD39:
**Please note:** I have zero experience in Go, so please forgive me if I haven't followed/done things correctly.

This pull requests fixes a few issues relating to podcast. They are:

- Adds the `artist` and `album` fields to the podcast episode response. There is a [pull request](https://github.com/sentriz/gonic/issues/565) opened relating to this.
- Returns podcast episodes in the `getBookmarks` response.
- I also added an `originalImageURL` for an episode, however, I am not sure if this is allowed as it is not part of the Subsonic API. My reason for adding this is because an episode could have its own image and I would prefer to use that rather than the main podcast image. Happy to revert if not possible. 

Example of episodes with its own image:


VD39:
@sentriz I would love to get your feedback on this when you get a minute. The test are failing, however, I don't know if they are related to the code. Thanks.

sentriz:
hi @VD39, CI should be fixed if you rebase from master

also I can't find any reference to `originalImageUrl`, is it documented somewhere?

VD39:
> hi @VD39, CI should be fixed if you rebase from master
> 
> also I can't find any reference to `originalImageUrl`, is it documented somewhere?

@sentriz the pipeline has been successfully fixed. Regarding the `originalImageUrl`, are you referring to its documentation in the Subsonic API? If so, it's not currently included there. I added this feature as an enhancement beyond the base API, as I believe displaying the original episode image is preferable to using the podcast's main image. Since this project is built on the Subsonic API, I'm unsure if making such additions or modifications is permissible.

sentriz:
> > hi @VD39, CI should be fixed if you rebase from master
> > also I can't find any reference to `originalImageUrl`, is it documented somewhere?
> 
> @sentriz the pipeline has been successfully fixed. Regarding the `originalImageUrl`, are you referring to its documentation in the Subsonic API? If so, it's not currently included there. I added this feature as an enhancement beyond the base API, as I believe displaying the original episode image is preferable to using the podcast's main image. Since this project is built on the Subsonic API, I'm unsure if making such additions or modifications is permissible.

ah okay got it. in that case, how about removing it from this PR. but and propose the new field [here](https://github.com/opensubsonic/open-subsonic-api/discussions)

then we can merge this early

VD39:
> > > hi @VD39, CI should be fixed if you rebase from master
> > > also I can't find any reference to `originalImageUrl`, is it documented somewhere?
> > 
> > 
> > @sentriz the pipeline has been successfully fixed. Regarding the `originalImageUrl`, are you referring to its documentation in the Subsonic API? If so, it's not currently included there. I added this feature as an enhancement beyond the base API, as I believe displaying the original episode image is preferable to using the podcast's main image. Since this project is built on the Subsonic API, I'm unsure if making such additions or modifications is permissible.
> 
> ah okay got it. in that case, how about removing it from this PR. but and propose the new field [here](https://github.com/opensubsonic/open-subsonic-api/discussions)
> 
> then we can merge this early

@sentriz I will proceed with the removal. Regarding the proposal for a new field, it seems unnecessary since the episode already has a [coverArt field](https://github.com/sentriz/gonic/pull/569/files#diff-cbb09452da9d6ce684f37811c70331ed7d369292aae7bf75eb58e1fed1bcc264R36). The existing issue is that this field is populated with the `id` of the podcast instead of the actual episode art. I'm uncertain about the effort required to download and use the episode's image for the `coverArt` instead.

PR #570, opened 2025-03-03
Add duration to getPlaylists request.

VD39:
**Please note:** I have zero experience in Go, so please forgive me if I haven't followed/done things correctly.

When calling the `getPlaylists`, it doesn't return the `duration` of each playlist, it always returns `0`. I hope this is a way to fix that issue.

### DIFF
--- a/server/ctrlsubsonic/handlers_playlist.go
+++ b/server/ctrlsubsonic/handlers_playlist.go
@@ -205,9 +205,6 @@ func playlistRender(c *Controller, params params.Params, playlistID string, play
 		Public:    playlist.IsPublic,
 		Owner:     user.Name,
 	}
-	if !withItems {
-		return resp, nil
-	}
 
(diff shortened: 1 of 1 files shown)


server/ctrlsubsonic/handlers_playlist.go
```diff
@@ -238,10 +235,14 @@ func playlistRender(c *Controller, params params.Params, playlistID string, play
...
 		trch.TranscodeMeta = transcodeMeta
-		resp.List = append(resp.List, trch)
+		if withItems {
+			resp.List = append(resp.List, trch)
+		}
 	}
 
-	resp.SongCount = len(resp.List)
+	if withItems {
+		resp.SongCount = len(resp.List)
```
khers, line 244: Your commit log mentions duration, but you are only filling out the SongCount field, is this intentional?

sentriz:
no activity since @khers commend, and no mention of duration - closing for now

PR #573, opened 2025-03-26
Fix transcoding typos

grilix:
This fixes a typo on a log message, and another on a function name.

Heads up for the log message: although I think it's better to have it correctly, it would impact anyone using this downstream.
For example, I found this while creating a filter on grafana to see how often transcodings were happening. For anyone who has already setup, either something similar or anything that relies on that log message, they'll have to update their configuration.

I don't know if that's even happening, since maybe that would have resulted on this same change before, but wanted to make sure we are aware of that.


sentriz (approved):

sentriz:
haha thanks, no idea how i missed that

PR #574, opened 2025-04-14
Use podcast author if episode author is not defined.

VD39:
This will use the podcast author if the podcast episode author is `undefined`. Should fix the issue stated in https://github.com/sentriz/gonic/issues/565.

### DIFF
--- a/db/migrations.go
+++ b/db/migrations.go
@@ -76,6 +76,7 @@ func (db *DB) Migrate(ctx MigrationContext) error {
 		construct(ctx, "202501152035", migrateTrackAddIndexOnAlbumID),
 		construct(ctx, "202501152036", migrateAlbumAddIndexOnParentID),
 		construct(ctx, "202502012036", migratePodcastEpisode),
+		construct(ctx, "2025040132036", migratePodcast),
 	}
 
 	return gormigrate.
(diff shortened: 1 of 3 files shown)


podcast/podcast.go
```diff
@@ -223,33 +224,41 @@ func (p *Podcasts) isAudio(rawItemURL string) (bool, error) {
...
 		Description: item.Description,
 		Title:       item.Title,
 		Length:      duration,
 		Size:        size,
 		PublishDate: item.PublishedParsed,
 		AudioURL:    audio,
 		Status:      db.PodcastEpisodeStatusSkipped,
-		Artist:      item.ITunesExt.Author,
-		Album:       podcastTitle,
+		Artist:      getPodcastEpisodeAuthor(item.ITunesExt.Author, podcast.Author),
```
sentriz, line 245:
```suggestion
		Artist:      cmp.Or(item.ITunesExt.Author, podcast.Author),
```

db/migrations.go
```diff
@@ -76,6 +76,7 @@ func (db *DB) Migrate(ctx MigrationContext) error {
 		construct(ctx, "202501152035", migrateTrackAddIndexOnAlbumID),
 		construct(ctx, "202501152036", migrateAlbumAddIndexOnParentID),
 		construct(ctx, "202502012036", migratePodcastEpisode),
+		construct(ctx, "2025040132036", migratePodcast),
```
sentriz, line 79: seems you have an extra digit here
```suggestion
		construct(ctx, "2025 04 132036", migratePodcast),
```

podcast/podcast.go
VD39: I was trying to find some documentation on the best way to do this, as `||` didn't work. Thanks for that.

VD39:
@sentriz requested changes have been made.

sentriz:
thanks!

Issue #578, opened 2025-05-11
createBookmark allows a user to create bookmark for invalid media ID.


khers:
gonic version: 0.16.4

if from docker, docker tag: stable

As a non-admin user, I can create a bookmark just fine, but when I attempt to retrieve the bookmarks for that user, I get an error return saying record not found.

```
>>> from libopensonic import Connection
>>> conn = Connection('https://music.example.com', username='user', password='password', port=443, server_path='/gonic')
>>> conn.create_bookmark('pe-181', 10900)
True
>>> conn.get_bookmarks()
{'subsonic-response': {'status': 'failed', 'version': '1.15.0', 'type': 'gonic', 'serverVersion': '0.16.4', 'openSubsonic': True, 'error': {'code': 10, 'message': 'finding entry: record not found'}}}
Traceback (most recent call last):
  File "<python-input-2>", line 1, in <module>
    conn.get_bookmarks()
    ~~~~~~~~~~~~~~~~~~^^
  File "/home/ebm/music-assistant/py-opensonic/src/libopensonic/connection.py", line 940, in get_bookmarks
    self._check_status(dres)
    ~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/ebm/music-assistant/py-opensonic/src/libopensonic/connection.py", line 2424, in _check_status
    raise exc(result['error']['message'])
libopensonic.errors.ParameterError: finding entry: record not found
```

khers:
After reading the code for bookmark management, what I suspect is happening is that I have created a bookmark for an invalid ID. This was permitted by the CreateBookmark endpoint, but the GetBookmarks method tries to validate that the media exists.

Unfortunately, I don't have a way to remove the offending values but moving forward I think the create endpoint should validate the id provided.

PR #580, opened 2025-05-27
Skip bookmarks for missing items instead of failing

khers:
Currently if a bookmark exists for an item that does not exist in the db, either because it has been removed or it never did, the user is prevented from retrieving any bookmarks. We can be in this state if media has moved or been deleted, but also because the createBookmark endpoint does not verify the item id provided.

This patch makes the getBookmarks endpoint ignore bookmarks for missingitems.

Fixes: https://github.com/sentriz/gonic/issues/578

### DIFF
--- a/server/ctrlsubsonic/handlers_bookmark.go
+++ b/server/ctrlsubsonic/handlers_bookmark.go
@@ -45,7 +45,12 @@ func (c *Controller) ServeGetBookmarks(r *http.Request) *spec.Response {
 				Find(&track, "id=?", bookmark.EntryID).
 				Error
 			if err != nil {
-				return spec.NewError(10, "finding entry: %v", err)
+				/*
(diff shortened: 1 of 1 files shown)


sentriz:
> I can also see cleaning the bookmarks for missing items at this point as well, happy to add that if it is the preferred solution.
> 
> Not returning them and not cleaning them means that this table can grow indefinitely.

@khers makes sense, we could probably have some cleanBookmarks in the scanner
image

khers:
> > I can also see cleaning the bookmarks for missing items at this point as well, happy to add that if it is the preferred solution.
> > Not returning them and not cleaning them means that this table can grow indefinitely.
> 
> @khers makes sense, we could probably have some cleanBookmarks in the scanner image

I will look into this but it might take me a minute, totally new to go.

PR #586, opened 2025-07-12
Add support for OIDC authentication

lava:
# Motivation

Passwords are old-school, and many people who are into self-hosting their software already have a centralized identity server for SSO. By offloading authentication to a separate service, we immediately gain support for 2FA, password reset flows, passkey support, identity federation, and whatever else your identity provider of choice offers.

This PR goes a bit beyond the minimal possible OIDC support, but in exchange it gets to a state where all the existing behavior can still be used without disruption when opting for OIDC.

In particular:
 - Allows users to choose between `password`, `oidc` and `oidc-forward` authentication methods. By default, `password` is selected and the gonic behavior is exactly as it was before this PR.
 - When choosing OIDC authentication, the username is read from the JWT provided by the identity provider, and users are created automatically when they are first seen. A new db field is added to allow users to freely change their name and still be identified as the same user by the `sub` claim.
- For new OIDC users a random password is generated and can be revealed in the UI, which can then be used for subsonic api clients. OIDC support for authorizing API access would make sense eventually, but is not part of this PR.
- The two different flavors `oidc` and `oidc-forward` correspond to the two popular choices of deploying OIDC: one is that the app natively supports OIDC and redirects users to the identity provider itself, the other is that something external to the app (like oauth2-proxy) is handling the whole login and just forwards the obtained token which the app then reads. Which one is preferred depends heavily on the other software in use, and in the end they share 95% of the same code, so I just added support for both.
- The check for admin-ness is a bit half-baked (only works by checking the `roles` claim in the token), but works for an initial implementation. 

# AI disclaimer

I'm not a Go developer, so most of the actual code was written by Claude. I did however work professionally on implementing OIDC support for various products, and looked over all the generated code to verify it's correct to the best of my ability.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ dist
 *.swp
 .tags*
 *.test
+data/
--- a/README.md
+++ b/README.md
@@ -73,7 +73,49 @@ password can then be changed from the web interface
 | `GONIC_MULTI_VALUE_ALBUM_ARTIST` | `-multi-value-album-artist` | **optional** setting for multi-valued album artist tags when scanning ([see more](#multi-valued-tags-v016))                                                                                                                                                                       |
 | `GONIC_TRANSCODE_CACHE_SIZE` | `-transcode-cache-size` | **optional** size of the transcode cache in MB (0 = no limit) |
 | `GONIC_TRANSCODE_EJECT_INTERVAL` | `-transcode-eject-interval` | **optional** interval (in minutes) to eject transcode cache (0 = never) |
+| `GONIC_AUTH_METHOD`              | `-auth-method`              | **optional** authentication method: `password` (default), `oidc`, or `oidc-forward` ([see more](#oidc-authentication))                                                                                                                                                           |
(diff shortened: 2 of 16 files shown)


Akruidenberg:
Very nice to have!

ChlorideCull:
This would be nice to have, but a few obvious things:

1. You do actually need to validate state to prevent CSRF.
2. Is there really no library that can handle the OIDC intricacies? This implementation isn't spec compliant (lenient with `use` on keys in the JWKs document, does not refresh the JWKs document when encountering an unknown `kid`) and adds a lot of security critical code.

snoweuph:
> Is there really no library that can handle the OIDC intricacies?

There is [`github.com/coreos/go-oidc/v3/oidc`](https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc) from [fedora coreOS](https://fedoraproject.org/coreos/) which is based on top of [`golang.org/x/oauth2`](https://pkg.go.dev/golang.org/x/oauth2)

Used it at work a bit to create an internal tool a few months ago.

PR #604, opened 2025-09-23
feat(build): Cgo free by default

sentriz:
- [x] tag backend changed to [wrtag](https://github.com/sentriz/wrtag) (Wasm based, pre normalised)
- [x] database changed to [ncruces/go-sqlite3](https://github.com/ncruces/go-sqlite3) (Wasm based)
- [x] Build flag to revert back to Cgo `-tags nowasm`
- [x] Remove taglib-dev sqlite-dev deps from CI
- [x] Cross compile without QEMU in CI
- [ ] Probably replace Cgo taglib with ffprobe reader (?)

---

image


sammy-ette:
why not the [modernc sqlite package](https://pkg.go.dev/modernc.org/sqlite)? TIL there is a wasm based sqlite package

sentriz:
> why not the [modernc sqlite package](https://pkg.go.dev/modernc.org/sqlite)? TIL there is a wasm based sqlite package

that's a nice module but the performance [isn't there](https://github.com/cvilsmeier/go-sqlite-bench), plus I have a lot more trust in an official sqlite Wasm binary running in a Go sandbox (wazero) compared to custom libc and C to Go transpiler. At least with regards to correctness and security

Issue #89, opened 2020-08-25
Support for multiple albums in one folder


Thor77:
In some cases, I store music files from different albums in one folder.
It looks like that structure is not really usable with gonic, because it just deletes songs which don't have the same artist as the album associated with the folder they're in (at least that what the behaviour of `Scanner.cleanArtists` looks like to me).
Additionally gonic will only pick up the album tag of one of the files in the folder, which is also very.. unfortunate for my usecase.

Is there a reason it is implemented this way? Are there plans to change it/would you accept a PR doing that?

sentriz:
hey sorry gonic isn't working out for you. 

i made this decision as compromise to make both the subsonic tags and folders APIs do-able. it also solves an issue where an artist has [multiple albums with the same name](https://en.wikipedia.org/wiki/American_Football_(band)#Discography). where the only real way to distinguish them (without looking different combinations of tags) is by folder

raffomania:
could we document the folder structure gonic expects in the README? I've just found this after wondering why gonic doesn't work with my playlist folders :D

icyphox:
ah, that would explain why 90% of my music isn't getting scanned! it's all like this (a lot of mp3s in the same folder):
![image](https://user-images.githubusercontent.com/17672379/101900318-482e5180-3bd5-11eb-9b75-a288dea78738.png)

agree with @raffomania, perhaps documenting what folder structure is expected would be great.

sentriz:
added a section to the readme :+1: 
https://github.com/sentriz/gonic#directory-structure

balthild:
Hello! I have a question: where should I put those musics without albums?

sentriz:
@balthild you cab just put then in any folder on the filesystem

sonbill:
Hello, i just start running gonic, i scanned and got 1 track, how i can play this track on web? or this is for backend API, thanks

sentriz:
[...]

PR #612, opened 2025-10-14
Folder name as Album [Fixes #611 and #89]

### LINKED ISSUE #611: Release name in scrobbles wrong
See https://github.com/sentriz/gonic/blob/e61e27815f854312554904e53e0915bc34e4a8c5/listenbrainz/listenbrainz.go#L56

This is wrong for Last.fm also.

drshajul:
This small change, makes the Folder name as the album name. 

But for users like me, where most songs in a folder may not have the same album, this is a blessing.
Also, for Various Artists folders, this should be great.
I understand that this will mean, if one folder contains one album, the folder name should match the album name.

Fixes #611 and #89, maybe more

### DIFF
--- a/scanner/scanner.go
+++ b/scanner/scanner.go
@@ -490,9 +490,8 @@ func populateAlbumEmbeddedCover(tx *db.DB, album *db.Album, track *db.Track, trp
 }
 
 func populateAlbum(tx *db.DB, album *db.Album, trags map[string][]string, modTime, createTime time.Time) error {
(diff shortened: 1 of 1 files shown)


sentriz:
sorry this is not what we want, album.TagTitle is for the title of the album from the tags. it's used to return the album title for browsing by tags. for browsing by folder album.RightPath is used 

if anything, maybe a setting to scrobble with the RightPath would make more sense, but not sure about that either

Issue #533, opened 2024-08-16
Missing getNowPlaying endpoint in API


dacagi:
gonic version: v0.16.4
docker-tag: latest

Not sure if an issue or a feature request. I just noticed gonic is missing the getNowPlaying endpoint based on the Subsonic API [specification](https://www.subsonic.org/pages/api.jsp#getNowPlaying).

I was looking to integrate gonic in my Homepage dashboard by reusing their existing Navidrome widget. I expected them to be almost compatible to show some basic stats.

I found the definitions for the endpoints [here](https://github.com/sentriz/gonic/blob/master/server/ctrlsubsonic/ctrl.go) and was surprised to find so many options but not getNowPlaying.

Is this something that was missed somehow or did you decide to not include that specific endpoint?

sentriz:
hi, thanks for report. its a dupe of #510

Issue #510, opened 2024-06-05
How to scrobble to both Listenbrainz and Maloja?


alscaldas:
Gonic version: sentriz/gonic:latest

Not a issue or bug, in fact... More of a question. I've just installed and configured Gonic and I'm trying to use its scrobbling features. Currently I scrobble to Lastfm, Listenbrainz and a selfhosted Maloja instance. But I noticed it seems to be possible to configure a single profile of each service. Is that so?

I also tried to configure gonic on a previously configured multi-scrobbler I also selfhost (which would be easier, since it already does everything I need), but it doesn't detect the songs played, although it appears on the multi-scrobbler dashboard. Is there any further configuration needed on gonic side to make it work this way?

sentriz:
(as discussed on IRC - updated title, so that "pull" scrobbers like multi-scrobber can work)

Issue #459, opened 2024-01-18
feat: song level play counts


sentriz:

aero31aero:
Hi, since you'd be coding this feature from scratch, for the database key, can you please use something like `acoustid_id` field if present, with the filepath as a fallback? That way, if a user finds out they need to modify some metadata, they can do it without losing track of their tracks.

I bet a lot of people running gonic already use some form of beets/picard or similar and have musicbrainz tags in their files.

sentriz:
it wouldn't be from scratch, just remodelling the schema to key by track,user instead of album,user 

acoustid or maybe mb recordingid is interesting though. I'll consider that

for my personal usecase it's not as important since my source of truth for playcounts is lastfm which are synced to my beets database (beets can also generate smart playlists that gonic can parse and serve)

aero31aero:
I do that too (`beet lastimport`) but I supposed it'd be way out of scope of gonic to parse beets database directly, hence my suggestion.

I like your workflow suggestion though. I could rely on beets smart playlists instead of symfonium's, since all I'm doing is maintaining offline downloads using them.

PR #632, opened 2025-12-04
Implement Subsonic getNowPlaying API endpoint

langelgjm:
This is a proof of concept implementation of the getNowPlaying API endpoint (see #510 ). The response NowPlayingEntry only provides the minimal fields required by the OpenSubsonic spec, plus "artist" and "album". The IsDir and and PlayerId fields are hardcoded to false / 0.

I implemented this so I could use [Homepage's Navidrome widget](https://gethomepage.dev/widgets/services/navidrome/) (which is just a generic Subsonic widget that hits the getNowPlaying endpoint and looks for Username, Title, Album, and Artist) as discussed in #533 .

Originally I had created a now_playing table in the database and would insert a row during ServeStream, but that seemed too heavy and invasive, so instead this code relies on an in-memory cache that simply stores the latest track streamed by each user.

Caveats:

- Since the cache is keyed by user, if the same user is streaming to multiple devices, we only see their most recent stream. Lots of ways to change this behavior, but I wanted to get feedback first.
- There's no cache eviction/expiration, but since it is keyed by user it can only grow as large as the number of users.
- Data is not persisted, which seems fine.
- I have not tested this with anything other than regular music tracks (no podcasts, internet radio, etc.)

Questions / Clarifications

- How do we feel about this implementation (cache vs. database table vs. some other method)?
- As far as I could tell, there's no way to get the data needed for getNowPlaying from the database; the plays table only provides album play counts.
- Should we aim to support as many of the optional fields of NowPlayingEntry as possible? Or keep it simple?
- Since getNowPlaying's NowPlayingEntry has a "minutesAgo" field, presumably the endpoint should return entries for recent plays even after the track has ended. Here, I limit returned entries to those within the last 60 minutes.

Full disclosure, I never used Go before this week and this is mostly Copilot code, but I have reviewed it, and I'm running this version of the code on my personal network right now.

### DIFF
--- a/server/ctrlsubsonic/ctrl.go
+++ b/server/ctrlsubsonic/ctrl.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"sync"
 	"time"
 
 	"go.senan.xyz/gonic/db"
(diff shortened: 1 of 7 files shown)


sentriz:
oh nice, thank you @langelgjm 

but i think we do in fact want to store track level plays in the database. that will be required for #459 

i'm not sure about the structure though. maybe extending the plays table, maybe a new table, should it be polymorhic so we can track podcast plays, internet radio plays, etc

also, if we have the concept of track level plays, we could delete the concept of album level plays, since the info is already there

langelgjm:
I just updated this to run a SQL query against the track_plays table instead of using an in-memory cache.

sentriz:
thanks @langelgjm ! hope you don't mind, i updated it to be more gorm-native, using new loading mechanism (so that we have rich data) and added test

langelgjm:
Not at all, thanks for cleaning it up!

PR #649, opened 2026-01-14
getSimilarSongs for artists and albums

### LINKED ISSUE #648: getSimilarSongs returns error when supplied with artist id
gonic version: `0.19.0`

docker tag: `latest`

```
https://music/rest/getSimilarSongs?s=XXX&c=Tempus&t=XXX&u=admin&v=1.15.0&f=json&id=ar-149&count=30
```
is requested from tempus, but gonic returns:
```json
{
  "subsonic-response": {
    "status": "failed",
    "version": "1.15.0",
    "type": "gonic",
    "serverVersion": "0.19.0",
    "openSubsonic": true,
    "error": {
      "code": 10,
      "message": "please provide an track `id` parameter"
    }
  }
}
```

however the use of `getSimilarSongs` for artists is correct, as per the [documentation](https://subsonic.org/pages/api.jsp#getSimilarSongs).

SnoutBug:
Closes #648 

`getSimilarSongs` now also provides tracks for artists and albums. For the former, code was taken from the already existing `getSimilarSongs2`.

The latter chooses a random track from the album on which to perform `track.getSimilar` as `last.fm` does not provide a corresponding `album.getSimilar`. The first successful response is taken.

I was also considering performing a `track.getSimilar` for all tracks, however, I think that this is unnecessary as tracks in an album are similar anyways, most of the times.

### DIFF
--- a/server/ctrlsubsonic/handlers_by_tags.go
+++ b/server/ctrlsubsonic/handlers_by_tags.go
@@ -664,19 +664,29 @@ func (c *Controller) ServeGetSimilarSongs(r *http.Request) *spec.Response {
 		return spec.NewError(10, "please provide an `id` parameter")
 	}
 
-	if id.Type == specid.Track {
-		return getSimilarSongsFromTrack(c, id, params, user, count)
+	var tracks []*spec.TrackChild
+	var sub *spec.Response
+
+	switch id.Type {
+	case specid.Track:
+		tracks, sub = getSimilarSongsFromTrack(c, id, params, user, count)
+	case specid.Album:
+		tracks, sub = getSimilarSongsFromAlbum(c, id, params, user, count)
+	case specid.Artist:
+		tracks, sub = getSimilarSongsFromArtist(c, id, params, user, count)
+	default:
+		return spec.NewError(10, "please provide a artist, album or track `id` parameter")
 	}
 
-	if id.Type == specid.Album {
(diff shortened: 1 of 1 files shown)


server/ctrlsubsonic/handlers_by_tags.go
```diff
@@ -787,6 +772,110 @@ func (c *Controller) ServeGetSimilarSongsTwo(r *http.Request) *spec.Response {
 	return sub
 }
 
+func (c *Controller) getSimilarSongsFromAlbum(id specid.ID, params params.Params, user *db.User, count int) *spec.Response {
```
sentriz: - can we move the getSimilarSongsXXX functions be placed after the ServeGetSimilarSongs and ServeGetSimilarSongsTwo handlers? 
 
- also instead of 

func (c *Controller) getSimilarSongsX(args)

can we make it

func getSimilarSongsX(c *Controller, args)

to match the style of other handler "helpers"

sentriz (commented):
thanks!

SnoutBug:
Thanks for the feedback.
I changed the return values for all `getSimilarSongsFrom` functions so that they are more in-line with another, since I had to change `getSimilarSongsFromArtist` anyways.

Happy to do any further adjustments.

sentriz:
thanks!

PR #650, opened 2026-01-16
Add playlist cover art support

D35P4C1T0:
- Add support for playlist cover art stored in user-specific covers folders
- Cover art should be placed in <playlists-path>/<user-id>/covers/<playlist-name>.jpg
- Supports both .jpg and .jpeg formats
- Cover art is cached and resized like other cover art
- Add BasePath() method to playlist.Store for cover lookup
- Add comprehensive tests for playlist cover art functionality

### DIFF
--- a/server/ctrlsubsonic/ctrl.go
+++ b/server/ctrlsubsonic/ctrl.go
@@ -61,6 +61,7 @@ type Controller struct {
 	cacheCoverPath  string
 	jukebox         *jukebox.Jukebox
 	playlistStore   *playlist.Store
+	playlistIDMap   *playlistIDMap
 	scrobblers      []scrobble.Scrobbler
 	podcasts        *podcast.Podcasts
 	transcoder      transcode.Transcoder
@@ -84,6 +85,7 @@ func New(dbc *db.DB, scannr *scanner.Scanner, musicPaths []MusicPath, podcastsPa
 		cacheCoverPath:  cacheCoverPath,
 		jukebox:         jukebox,
 		playlistStore:   playlistStore,
+		playlistIDMap:   newPlaylistIDMap(),
 		scrobblers:      scrobblers,
 		podcasts:        podcasts,
 		transcoder:      transcoder,
--- a/server/ctrlsubsonic/handlers_playlist.go
+++ b/server/ctrlsubsonic/handlers_playlist.go
@@ -1,7 +1,6 @@
 package ctrlsubsonic
 
 import (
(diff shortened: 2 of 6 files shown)


sentriz:
thanks for the PR! though not sure about the direction here

i would expect we could have just done

coverFor(... c.playlistsPath) 

which internaly checks

	case specid.Playlist:
		return coverForPlaylist()

of course, we would need to start generating playlist spec IDs from subsonic playlist endpoints, and recieving them in this endpoint

also the refactor the the endpoint (new functions / spliting /etc ) is hard to read, can we do without that?

thanks!

D35P4C1T0:
Thanks for the feedback! I could make so that playlists are treated as first-class specid entities. This adds specid.Playlist and uses deterministic IDs (i.e. derived from playlist paths), avoiding database changes while keeping playlist handling consistent with the rest of the API. For compatibility, playlist endpoints can still accept the existing base64 IDs.

With that in place, playlist cover art can be handled through the existing coverFor() switch via a specid.Playlist case, rather than separate logic. This should keep cover resolution unified and allows ServeGetCoverArt to stay simple.

I'm implementing this direction in the next version of the PR, but wanted to confirm the approach makes sense before finalizing it.

sentriz:
that sounds good! keep in mind, i haven't looked how easy it would be to make a specid.Playlist, but let's see 

> For compatibility, playlist endpoints can still accept the existing base64 IDs.

i don't think we need to care about this. most clients will getPlaylists then use that render a UI and use. I wouldn't expect any clients to cache playlist IDs (could be wrong on that though)

D35P4C1T0:
Thanks for the feedback, I’ve incorporated it here.

I think I handled the specid.Playlist implementation. I also updated all the playlist endpoints to use it. 
Main changes:
- Added specid.Playlist ("pl") to the specid package
- Added a playlistIDMap that maps playlist file paths to deterministic numeric specid.IDs using FNV-64a hashing
- Updated getPlaylists, getPlaylist, createPlaylist, updatePlaylist, and deletePlaylist to generate and accept specid.Playlist IDs
- Dropped backward compatibility for base64-encoded IDs as suggested
- Wired playlist cover art into the coverFor switch

sentriz:
hey, thanks for the PR - i merged it but with some changes,

- the playlist id map is not needed, we can store the path in the specid
- opted for the cover for x.m3u be x.png (instead of covers dir)
- using existing coverparse logic for that

PR #668, opened 2026-02-24
Fix crash if itunes extensions are missing

diepfote:
I came across this issue when I started to use a custom RSS generator that works fine for e.g. the Apple Podcasts app but results in a crash in gonic.

I found these nil pointer dereferences by looking at the stacktraces only.
No time was spent looking for additional locations were this might need to be fixed.

The custom generator if that helps
https://gist.github.com/diepfote/59aaf5eb8ea3e2d7bcdd97b8efd2a472

### DIFF
--- a/podcast/podcast.go
+++ b/podcast/podcast.go
@@ -81,13 +81,17 @@ func (p *Podcasts) AddNewPodcast(rssURL string, feed *gofeed.Feed) (*db.Podcast,
 	if err != nil {
 		return nil, fmt.Errorf("find unique podcast dir: %w", err)
 	}
+	var author string
+	if feed.ITunesExt != nil {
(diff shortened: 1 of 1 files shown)


diepfote:
[Here](https://git.sr.ht/~j_wernick/localhost-podcast/tree/37f4014a4d6b3942de18767c6e0a916edf7f6b52/item/src/rss_feed.rs) is where the generator uses the rss crate. rss::ChannelBuilder would need a [itunesExtension](https://github.com/rust-syndication/rss/blob/7d9627a92e93632a6458970b353c440195fc43eb/src/channel.rs#L101) parameter to output these.

As it seems to be an extension I'd consider expecting it to be there  a bug.

sentriz:
thanks!

PR #687, opened 2026-04-27
feat: add cache size and eject interval for cover art cache

brian-doherty:
## Summary

- Adds `-cover-cache-size` (MB) and `-cover-eject-interval` (minutes) flags, mirroring the existing `-transcode-cache-size` / `-transcode-eject-interval` flags
- Introduces `CoverCache` in the `ctrlsubsonic` package, holding the cache path, size limit, and a `sync.RWMutex` to prevent races between cover writes and cache ejection
- `ServeGetCoverArt` holds a read lock for its entire execution; `CacheEject` takes the exclusive write lock before deleting files
- Touches cached cover files on access so LRU ordering is accurate when the eject runs

## Test plan

- [x] Start gonic with `-cover-cache-size 10 -cover-eject-interval 1` and verify covers are served correctly
- [x] Confirm the cache directory is pruned to the size limit after the eject interval elapses
- [x] Verify that omitting either flag (defaulting to 0) disables ejection, matching transcode behaviour
- [x] `go test ./...` passes
- [x] `golangci-lint run` passes

### DIFF
--- a/server/ctrlsubsonic/covercache.go
+++ b/server/ctrlsubsonic/covercache.go
@@ -1,0 +1,66 @@
+package ctrlsubsonic
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"sync"
+)
+
+type CoverCache struct {
+	path      string
+	limitMB   int
+	cleanLock sync.RWMutex
+}
+
+func NewCoverCache(path string, limitMB int) *CoverCache {
+	return &CoverCache{path: path, limitMB: limitMB}
+}
+
+func (c *CoverCache) CacheEject() error {
+	c.cleanLock.Lock()
+	defer c.cleanLock.Unlock()
+
+	type file struct {
+		path string
+		info os.FileInfo
+	}
+
+	var files []file
+	var total int64
+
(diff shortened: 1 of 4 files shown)


brian-doherty:
@sentriz For your consideration -- cover cache eject to match the transcode cache eject, for users running under disk space constraints.

brian-doherty:
If this goes in, what I'm thinking about next is pre-caching the cover for albums or songs returned from a search call, under the assumption that they're likely about to be requested, but I'd love your thoughts on this.

server/ctrlsubsonic/covercache.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"sync"
+)
+
+type CoverCache struct {
```
sentriz, line 12: @brian-doherty this seems to be a useful type. i wonder would it make sense to drop the Cover name and re use it for transcode cache?

(if that makes sense, I'm on phone atm)

brian-doherty: That's a good idea. Do you want to accept this PR and have me make that the next change? Or make it all one PR?

sentriz: shipped, follow up sounds good. thanks!

PR #688, opened 2026-05-02
feat: add ISRC to child data

### LINKED ISSUE #642: Retrieve ISRC from file tags
https://github.com/opensubsonic/open-subsonic-api/pull/148 has been merged, so it would be cool to start returning the ISRC(s) of a track if it exists in its metadata.

I've started working on a PR for this, but opening an issue now just in case there are any objections/tips in the meantime.

gabeklavans:
spec: https://github.com/opensubsonic/open-subsonic-api/pull/148

I heavily based its design on genres, since they're essentially treated the same way.

I added tests for it in scanner, let me know if I should add tests anywhere else.

closes #642 (without the ability to search by ISRC, but I can get to that later)


db/db.go
```diff
@@ -429,15 +435,18 @@ const (
...
-	TrackID   int  `gorm:"not null; unique_index:idx_track_id_genre_id" sql:"default: null; type:int REFERENCES tracks(id) ON DELETE CASCADE"`
-	GenreID   int  `gorm:"not null; unique_index:idx_track_id_genre_id" sql:"default: null; type:int REFERENCES genres(id) ON DELETE CASCADE"`
-	Inherited bool `gorm:"not null; default:false"`
+	TrackID int `gorm:"not null; unique_index:idx_track_id_genre_id" sql:"default: null; type:int REFERENCES tracks(id) ON DELETE CASCADE"`
+	GenreID int `gorm:"not null; unique_index:idx_track_id_genre_id" sql:"default: null; type:int REFERENCES genres(id) ON DELETE CASCADE"`
+}
+
+type TrackISRC struct {
+	TrackID int `gorm:"not null; unique_index:idx_track_id_isrc_id" sql:"default: null; type:int REFERENCES tracks(id) ON DELETE CASCADE"`
+	IsrcID  int `gorm:"not null; unique_index:idx_track_id_isrc_id" sql:"default: null; type:int REFERENCES isrcs(id) ON DELETE CASCADE"`
```
sentriz: @gabeklavans i wonder is there any benefit to having a new table to hold ISRCs? they are aready an id themselves. wdyt about having this col just be the isrcs?

scanner/scanner.go
```diff
@@ -937,6 +935,24 @@ func (s *Scanner) cleanGenres(st *State) error { //nolint:unparam
 	return nil
 }
 
+func (s *Scanner) cleanISRCs(st *State) error {
```
sentriz: i suppose this wouldn't be needed if we didnt have the new table, the ON CASCASE rules would clean up

db/db.go
gabeklavans: I added this table because a track can have multiple isrcs, and the spec requires the value to be an array in the return. If we had it as a single string column in the db, we'd have to do some processing to format it right, I think

gabeklavans: I agree it feels wrong to implement this in a "many to many" way, though, since an ISRC won't ever be used by more than one track. Do you have a better way to handle it in mind?

sentriz: that can still happen I just mean

```diff
-type ISRC struct {
-	ID    int    `gorm:"primary_key"`
-	Value string `gorm:"not null; unique_index"`
-}
-
 type TrackISRC struct {
 	TrackID int `gorm:"not null; unique_index:idx_track_id_isrc_id" sql:"default: null; type:int REFERENCES tracks(id) ON DELETE CASCADE"`
- 	IsrcID  int `gorm:"not null; unique_index:idx_track_id_isrc_id" sql:"default: null; type:int REFERENCES isrcs(id) ON DELETE CASCADE"` 
+ 	ISRC    int `gorm:"not null; unique_index:idx_track_id_isrc_id" sql:"default: null"` 
 }
```

where it becomes a string, a track can still have many ISRCs via multiple rows to the table

gabeklavans: oh.. good point, yeah.

gabeklavans:
Simplified according to the review comments.

I also removed the index creation in the migration. My intention was to put it there for when I add searching by ISRC, but it might not be necessary yet. Also the unique_index on TrackISRC might cover that? Not sure.

db/db.go
```diff
@@ -434,6 +435,11 @@ type TrackGenre struct {
 	Inherited bool `gorm:"not null; default:false"`
 }
 
+type TrackISRC struct {
+	TrackID int    `gorm:"not null unique_index:idx_track_isrc" sql:"default: null; type:int REFERENCES tracks(id) ON DELETE CASCADE"`
+	ISRC    string `gorm:"not null; unique_index:idx_track_isrc" sql:"default: null;"`
```
sentriz: @gabeklavans if we swap these (and isrc is just first column on the index) then searching by ISRC string will use the index 

```suggestion
	ISRC    string `gorm:"not null; unique_index:idx_isrc_track" sql:"default: null;"`
	TrackID int    `gorm:"not null unique_index:idx_isrc_track" sql:"default: null; type:int REFERENCES tracks(id) ON DELETE CASCADE"`
```

server/ctrlsubsonic/spec/spec.go
```diff
@@ -231,7 +231,8 @@ type TrackChild struct {
 	Type        string      `xml:"type,attr,omitempty"        json:"type,omitempty"`
 	Year        int         `xml:"year,attr,omitempty"        json:"year,omitempty"`
 
-	MusicBrainzID string `xml:"musicBrainzId,attr"        json:"musicBrainzId"`
+	MusicBrainzID string   `xml:"musicBrainzId,attr"        json:"musicBrainzId"`
+	ISRC          []string `xml:"isrc,omitempty"            json:"isrc,omitempty"`
```
sentriz: for opensubsonic fields, we usually don't to omitempty, so that clients can know we support the field

sentriz: instead in the spec.NewXfromY, you crate an empty non-nil slice like 

`ISRCs: []string{}`

then later loop and append to it

server/ctrlsubsonic/testdata/test_get_music_directory_album_aa_jamstash
```diff
@@ -53,6 +53,9 @@
           "type": "music",
           "year": 2018,
           "musicBrainzId": "00000000-0000-0000-0000-aa0000000000",
+          "isrc": [
+            ""
+          ],
```
sentriz: why the empty string here? i would have expected "isrc": [] i suppose

sentriz (commented):
thanks! just a couple more things

server/ctrlsubsonic/testdata/test_get_music_directory_album_aa_jamstash
sentriz: maybe we're insert empty isrcs into the db?

gabeklavans: hm yeah that’s a point. I’ll check on that

scanner/scanner.go
```diff
@@ -678,6 +683,24 @@ func populateTrackGenres(tx *db.DB, track *db.Track, directIDs, inheritedIDs []i
...
+func populateTrackISRCs(tx *db.DB, track *db.Track, isrcs []string) error {
+	if err := tx.Where("track_id=?", track.ID).Delete(db.TrackISRC{}).Error; err != nil {
+		return fmt.Errorf("delete old track ISRCs records: %w", err)
+	}
+
+	var col [][]any
+	for _, isrc := range isrcs {
+		if isrc == "" {
+			continue
+		}
```
gabeklavans, line 695: It seems that [my usage of normtag](https://github.com/sentriz/gonic/pull/688/changes#diff-7d1df5e49f35d124e1d892865498cde1749af35afbb0f0022499a4b1b09a91eaR496) returns `[""]` even when the track file doesn't have an ISRC tag at all. Let me know if there's a way I can parse the tag that would return an actual empty array, meaning I wouldn't need this empty string check.

sentriz:
thanks!

PR #690, opened 2026-05-06
feat: track client cover size preference and prefetch on album/song responses

brian-doherty:
## Summary

- Adds `ClientCoverSizePreference` table (`user_id`, `client`, `last_cover_size`) to record the pixel size each client last requested via `getCoverArt`
- On every `getCoverArt` call, the requested size is upserted into the table in the background (non-blocking, including cache hits)
- The `resp` wrapper in `New()` now fires a background goroutine after every successful API response that contains album or track cover IDs, populating the cover cache at that client's preferred size — so subsequent `getCoverArt` calls are served directly from cache
- Artist covers are excluded from prefetch (they're fetched from external Last.fm URLs, not local files)

## Test plan

- [ ] Make a `getCoverArt` request with a specific `size` param — confirm a row appears in `client_cover_size_preferences`
- [ ] Call `getAlbum` or `getAlbumList2` — confirm cover files appear in the cache dir at the stored size without a direct `getCoverArt` request
- [ ] Change the size on a subsequent `getCoverArt` call — confirm the preference updates and new prefetches use the new size
- [ ] Confirm existing tests still pass: `go test ./...`


brian-doherty:
Reopening on a dedicated branch.

PR #710, opened 2026-07-05
feat(subsonic): Sorting like file explorers

ummshsh:
Good day!

I thought it makes sense to have option to see folders sorted just like you see them in file explorers. Certainly makes my live easier =) 

### Problem

Gonic sorts folder listings by `right_path COLLATE NOCASE` — raw bytewise comparison. This means `(` sorts before `_` (ASCII 40 vs 95).  
File explorers (Finder, Explorer, Nautilus) use the Unicode Collation Algorithm, which gives `_` a lower weight than `(` — so `_1` comes before `(1992) One`.

### Solution

**Pre-computed sort key column** — same approach file explorers use under the hood.
linguistic (UCA-based) folder sorting with admin toggle

Screenshot 2026-07-05 at 18 17 56

#### What changed (10 files)

- **`db/db.go`** — Added `RightPathSortKey` field to `Album` model + new `LinguisticSorting` setting key
- **`db/migrations.go`** — Migration `202607051200` adds the column via `AutoMigrate` and backfills sort keys for all existing albums using `golang.org/x/text/collate`
- **`scanner/scanner.go`** — `populateAlbumBasics` always writes a UCA sort key from the directory basename during scan
- **`server/ctrlsubsonic/ctrl.go`** — Three helper methods (`sortRightPath`, `sortRightPathPrefixed`, `sortRightPathParentPrefixed`) that check the `linguistic_sorting` setting and return either `right_path_sort_key` or `right_path COLLATE NOCASE`
- **`server/ctrlsubsonic/handlers_by_folder.go`** — All 6 `ORDER BY` clauses switched to use helpers
- **`server/ctrlsubsonic/handlers_by_tags.go`** — Tag-mode artist album listing switched to use helper
- **`server/ctrladmin/handlers.go`** — Loads setting in `ServeHome`, added POST handler to save toggle
- **`server/ctrladmin/ctrl.go`** — Registered route, added `LinguisticSorting` to template data
- **`server/ctrladmin/adminui/pages/home.tmpl`** — New "sorting" block with dropdown (bytewise / linguistic)
- **`go.mod`** — Promoted `golang.org/x/text` to direct dependency

#### How it works

The sort key is always written during scanning regardless of the toggle. The toggle (`/admin/home` → **sorting** dropdown) only changes which column the `ORDER BY` uses at query time — zero-rebuild, instant switch. Migration backfills existing albums so there's no gap.

#### Migration behavior

On upgrade, `migrateAlbumRightPathSortKey`:
1. Adds `right_path_sort_key TEXT` column via `AutoMigrate`
2. Iterates all albums and sets the sort key from `right_path` using UCA
3. No downtime — column is nullable, next scan fills any newly added albums

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -170,9 +170,10 @@ func (db *DB) TransactionChunked(data []int64, cb func(*DB, []int64) error) erro
 type SettingKey string
 
 const (
-	LastFMAPIKey SettingKey = "lastfm_api_key" //nolint:gosec
-	LastFMSecret SettingKey = "lastfm_secret"
-	LastScanTime SettingKey = "last_scan_time"
+	LastFMAPIKey     SettingKey = "lastfm_api_key" //nolint:gosec
+	LastFMSecret     SettingKey = "lastfm_secret"
+	LastScanTime     SettingKey = "last_scan_time"
+	LinguisticSorting SettingKey = "linguistic_sorting"
 )
 
 func (db *DB) GetSetting(key SettingKey) (string, error) {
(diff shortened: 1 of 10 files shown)


ummshsh:
This approach targets only folders, I don't know if it makes sense to do the same for individual tracks.

PR #712, opened 2026-07-16
feat(subsonic): ogg transcode profiles

fanyx:
mostly out of self interest, but as long as custom transcode profiles are not implemented, might be nice for some people.

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -47,6 +47,10 @@ var (
 
 	Opus192 = NewProfile("audio/ogg", "opus", 192, `ffmpeg -v 0 -i <file> -ss <seek> -map 0:a:0 -vn -b:a <bitrate> -c:a libopus -vbr constrained -f opus -`)
 
+	Ogg    = NewProfile("audio/ogg", "ogg", 112, `ffmpeg -v 0 -i <file> -ss <seek> -vn -c:a libvorbis -b:a <bitrate> -f ogg -`)
+	Ogg192 = NewProfile("audio/ogg", "ogg", 192, `ffmpeg -v 0 -i <file> -ss <seek> -vn -c:a libvorbis -b:a <bitrate> -f ogg -`)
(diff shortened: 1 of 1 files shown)


transcode/transcode.go
```diff
@@ -47,6 +47,10 @@ var (
 
 	Opus192 = NewProfile("audio/ogg", "opus", 192, `ffmpeg -v 0 -i <file> -ss <seek> -map 0:a:0 -vn -b:a <bitrate> -c:a libopus -vbr constrained -f opus -`)
 
+	Ogg    = NewProfile("audio/ogg", "ogg", 112, `ffmpeg -v 0 -i <file> -ss <seek> -vn -c:a libvorbis -b:a <bitrate> -f ogg -`)
```
sentriz, line 50: 1. The new profiles aren't added to the UserProfiles map, so nobody can select them. Needs "vorbis": Vorbis, etc
  2. I think we should rename to Vorbis. Codec is libvorbis, so Vorbis/Vorbis192/Vorbis256 (keys vorbis, vorbis_192, vorbis_256). Ogg names the container
  3. Can we add -map 0:a:0?. Every sibling profile pins the first audio stream
  4. Arg order differs from siblings. Uses -vn -c:a ... -b:a vs the established -b:a ... -c:a. just a nit 
  5. needs gofmt to be ran
  7. why is 112 kbps set? instead of way 128?

sentriz:
hm actually why would anyone want to use vorbis over opus @fanyx it's quite an old and inferior codec at this point

transcode/transcode.go
fanyx: 1-5. all reasonable
6. 112kbps is the default when you don't specify any -q:a or -b:a on ffmpeg (defaults to -q:a2 which is 112kbps)

fanyx:
Ok i did some experiments with opus files for my specific use case.
Initially I didn't want to use it since you can't embed cover art into the files, but that's fine on second thought.
Just need to adjust my app to save the art as cover.jpg in the corresponding folder then.

Guess we don't need vorbis transcode or users could add it once custom transcode profiles are implemented during runtime

PR #717, opened 2026-08-01
feat(ffprobe): Consider stream tags

### LINKED ISSUE #716: Tags are sometimes ignored
gonic version: 6ddbbfa0dfb87cbf6fc9323a26adcdca0b52845c, built with `-tags nowasm`

gonic does not use all available tags in my music library. I'm using the `/rest/search3` endpoint to retrieve songs. Many songs return with `Unknown Artist`, `Unknown Album` and the filename as title, instead of using the available tags.

I have used the `mediainfo` tool to verify, that the tags are there. The tool shows the correct info in the "Performer", "Album", "Track name/Position" and "Title" fields.

gonic's ability to use the tags also seems to depend on the file format. In an MPEG-4/AAC file, all tags except the album were correctly used by gonic, but on an Ogg/Opus file, not a single tag was used.

codesoap:
In my library, some songs show their artist, album, title and track number only in the stream tags, not in the `format` section of `ffprobe`. With this fix, the stream tags are considered, but the `format` tags would still overwrite the stream tags, if they are available additionally.

If there are multiple streams with stream tags, things will probably get ugly, but at least in my library, this doesn't seem to occur.

Fixes #716 (even the ignored album; I don't quite understand how - was it because problematic and unproblematic songs were mixed in one directory?!)

### DIFF
--- a/tags/ffprobe/ffprobe.go
+++ b/tags/ffprobe/ffprobe.go
@@ -49,21 +49,18 @@ func (Reader) Read(absPath string) (tags.Properties, tags.Tags, error) {
 	bitRateBitsPerSec, _ := strconv.Atoi(d.Format.BitRate)
 
 	var tgs = map[string][]string{}
-	var hasCover, tgsFound bool
+	var hasCover bool
 	for _, s := range d.Streams {
 		switch s.CodecType {
 		case "video":
 			hasCover = true
 		case "audio":
-			if !tgsFound {
-				tgsFound = true
-				for k, vs := range s.Tags {
-					tgs[k] = strings.Split(vs, ";")
-				}
+			if len(tgs) > 0 {
+				continue // first audio stream wins
 			}
-		}
-		if hasCover && tgsFound {
-			break
+			for k, vs := range s.Tags {
+				tgs[k] = strings.Split(vs, ";")
+			}
 		}
 	}
 	for k, vs := range d.Format.Tags {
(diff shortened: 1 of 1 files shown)


tags/ffprobe/ffprobe.go
```diff
@@ -48,6 +49,11 @@ func (Reader) Read(absPath string) (tags.Properties, tags.Tags, error) {
 	bitRateBitsPerSec, _ := strconv.Atoi(d.Format.BitRate)
 
 	var tgs = map[string][]string{}
+	for _, stream := range d.Streams {
```
sentriz: @codesoap i think we want to capture stream.Tags in the audio stream branch, then apply format tags on top. so that format still wins

this would fix two issues where
cover art streams's info leaks into the tag map, and also the non deterministic tag reading. this can be issue for multiple scans

codesoap: I have now added the filter for the `codec_type`. Is this what you had in mind?

Although I don't see, how this makes the code more deterministic :thinking:

sentriz: ah my mistake, d.Streams is a slice

the only thing left then: with multiple audio streams they still merge, last one winning. how about we take only the first audio stream's tags?

codesoap: Done.

